PR #25, opened 2021-12-08
Update README and tutorials

apeyrache:


gviejo (changes requested):
In main2_neuroseries, the namespace is still ap. Need to be changed from ap.Ts to nap.Ts
In main4_raw_data, nap.loadSpikeData does not return shank anymore. See main.py in main folder.
Same for main5.

Issue #49, opened 2022-01-21
decode without occupancy normalization


dlevenstein:
decode__1d and decode_2d should allow for feature=None, for conditions where you don't have, or don't want to bias by behavioral occupancy for decoding.

For example:
-sleep
-an open field experiment, where tuning curves are extracted from random foraging (full occupancy), but then experiment is done on navigation to remembered goal locations (limited occupancy)

gviejo:
Solved in 9d8ad62e8d46545351472cdb15e8af338ef91907

Issue #45, opened 2022-01-20
TsdFrame.value_from


dlevenstein:
Should TsdFrame have a value_from function, to match Tsd? 

(copying it over for continuous tuning curve function - solved on my fork will pull it in with the continuous tuning stuff)

Issue #19, opened 2021-12-06
Spatial Info and Tuning Curves (continuous and spike data)


dlevenstein:
Hey - looking to compute spatial info and place fields for continuous time data (in this case, activations from a simulated neural network, but could also work for e.g. calcium data). Does this already exist in pynapple? If not, will make the functions.

Looks like computeSpatialInfo is only for a 1D angular variable (HD) and computePlaceFields is for spikes, right?

gviejo:
Hey Dan, spatial_info is only for HD right now in functions.py and hasn't been tested yet within pynapple.
So far, computePlaceFields works with spikes but feel free to work on the function for continous data. If you want some inspirations, you can take a look [here](https://github.com/gviejo/PSBImaging/blob/main/python/functions.py). It is my repository to work with calcium data and it takes continous data as input.

dlevenstein:
Functiosn: 1D Tuning Curve, 2D tuning curve, ND tuning curve.

East: for continuous (tsd) or spike (ts) data

dlevenstein:
as part of this, ran into #29 re: bin standardization

Related: these functions have nb_bins as a scalar input, which produces a square shaped discretization of space. However, many spaces might be longer than they are wide. Maybe we should have the option to put in a (length, width) touple, which if an integer nb_bins is passed in is converted to (nb_bins, nb_bins)?

dlevenstein:
I noticed computePlaceFields saves place_fields as a dictionary of pd.DataFrames, would this be better as a pd.Panel?

gviejo:
It says Deprecated since version 0.20.0. I would suggest return a dictionnary then. Or eventually a 3d array.

dlevenstein:
Ah bummer. Dictionary probably better... to maintain cellID labels (similar to TSGroup)

gviejo:
Added in 9d8ad62e8d46545351472cdb15e8af338ef91907

PR #50, opened 2022-01-21
2d tuning curves and decoding with continuous data (TsdFrame)

dlevenstein:
-added compute_2d_tuning_curves_continuous (#19)

-added TsdFrame.value_from (#45)

modified decode_d2:
-take TsdFrame input for continuous data
-allow feature=None for no occupancy normalization (#49)
-annoying tab difference I can't get rid of 🙄


dlevenstein:
Note - I'll have to bring things over to the 1d cases, but figured I'd pull these for any feedback in the meantime

pynapple/process/decoding.py
```diff
@@ -134,64 +134,81 @@ def decode_2d(tuning_curves, group, feature, ep, bin_size, xy):
...
-		bins = np.hstack((bins, [bins[-1]+diff[-1],bins[-1]+2*diff[-1]])) # assuming the size of the last 2 bins is equal
-		binsxy.append(bins)
-
-	occupancy, _, _ = np.histogram2d(
-		feature.iloc[:,0],
-		feature.iloc[:,1],
-		[binsxy[0], binsxy[1]])
-	occupancy = occupancy.flatten()
+    if feature is None:
+        #For conditions (e.g. sleep) where you don't have, or don't want to
```
gviejo, line 165: Ok this is problematic here. Features is passed only to compute occupany in this function. It could be passed as an optional argument but then you need to pass it as occupancy = None in the function definition. 
Also is there a case where you don't want to correct for occupancy?

pynapple/process/decoding.py
```diff
@@ -134,64 +134,81 @@ def decode_2d(tuning_curves, group, feature, ep, bin_size, xy):
...
-		feature.iloc[:,0],
-		feature.iloc[:,1],
-		[binsxy[0], binsxy[1]])
-	occupancy = occupancy.flatten()
+    if feature is None:
+        #For conditions (e.g. sleep) where you don't have, or don't want to 
+        #bias by behavioral occupancy for decoding
+        occupancy = np.ones(1)
+        #TODO: Assuming features are x and y.... this is hacky
+        feature = nap.TsdFrame(1,columns = ('x','y'))
```
gviejo, line 169: I am missing the idea here :thinking:

pynapple/process/tuning_curves.py
```diff
@@ -123,6 +123,70 @@ def compute_2d_tuning_curves(group, feature, ep, nb_bins, minmax=None):
...
+        else:
+            bins = np.linspace(minmax[i+i%2], minmax[i+1+i%2], nb_bins)
+        binsxy[c] = bins
+        groups_value[c] = np.digitize(groups_value[c],binsxy[c])-1
+        # This might be better done with 
+        #groups_value[c] = pd.cut(groups_value[c],binsxy[c])
+        # to keep groups_value as a Tsd object, but that threw an error
+
+    tc = {}
+    tc_np = np.zeros((group.shape[1], nb_bins, nb_bins))
```
gviejo, line 177: It should be 
```
    tc_np = np.zeros((group.shape[1], nb_bins-1, nb_bins-1))
```
Otherwise you have one more bin.

pynapple/process/decoding.py
```diff
@@ -134,64 +134,81 @@ def decode_2d(tuning_curves, group, feature, ep, bin_size, xy):
...
+    decoded = nap.TsdFrame(
 		t = count.index.values,
 		d = np.vstack((xy[0][idxmax2d[0]], xy[1][idxmax2d[1]])).T,
 		time_support = ep,
 		columns=feature.columns
 		)
 
-	return decoded, p
\ No newline at end of file
+    return decoded, p
```
gviejo, line 214: I think it should work but can you provide with a minimal working example with toy data in your PR that can generate some figures to have a full check.

gviejo (changes requested):
Hey Dan, I wrote some comments about your PR.

dlevenstein:
Thanks - I'll be able to take a look at this tomorrow afternoon!

pynapple/process/decoding.py
dlevenstein: E.g. during sleep. You have to assume flat occupancy because there's no behavior.

pynapple/process/decoding.py
dlevenstein: I don't have any continuous neural data, but could provide an example with neural network data that makes (albeit still poorly) place fields and decoding? 😅

pynapple/process/decoding.py
dlevenstein: It was a hack so that the column labels got passed through at the end. Maybe the better thing to do would be if there's no Feature passed in, it replaces with a fake feature TsdFrame with uniform occupancy?

pynapple/process/tuning_curves.py
dlevenstein: It had the right number of bins when I ran it, will double check why

pynapple/process/tuning_curves.py
```diff
@@ -123,6 +123,70 @@ def compute_2d_tuning_curves(group, feature, ep, nb_bins, minmax=None):
     
     return tc, xy
 
+def compute_2d_tuning_curves_continuous(group, feature, ep, nb_bins, minmax=None):
+    """
+    Computes 2-dimensional tuning curves relative to a 2d feature
+    
+    Parameters
+    ----------
+    group: Tsd or TsdFrame or dict of Tsd objets
```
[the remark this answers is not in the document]
dlevenstein: fixed

pynapple/process/tuning_curves.py
```diff
@@ -123,6 +123,70 @@ def compute_2d_tuning_curves(group, feature, ep, nb_bins, minmax=None):
     
     return tc, xy
 
+def compute_2d_tuning_curves_continuous(group, feature, ep, nb_bins, minmax=None):
```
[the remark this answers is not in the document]
dlevenstein: fixed

gviejo (approved):

pynapple/process/decoding.py
gviejo: Ok let's give the option. If occupancy is none, it's flat.

PR #71, opened 2022-02-13
Read neuroscope intervals function

GRVite:
Adding a function to read .evt files. It first searches if they are present in the nwb file. If the events are not in the nwb, it loads them from the .evt files.  I had to modify load_nwb_intervals to make it work (now it checks if the interval is among in the nwb file before trying to load it).


pynapple/io/neurosuite.py
```diff
@@ -372,6 +372,35 @@ def load_lfp(self, filename=None, channel=None, extension='.eeg', frequency=1250
...
+        -------
+        IntervalSet  
+    
+        """
+        isets = self.load_nwb_intervals(name)
+        if str(type(isets)) !="<class 'pynapple.core.interval_set.IntervalSet'>":
+            print("reading intervals...")
+            evt_file = os.path.join(self.path, self.path.split('/')[-1] + "." + name + ".evt")
+            df = pd.read_csv(evt_file, delimiter=' ', usecols = [0], header = None)
+            start = nap.Ts(df.iloc[::2].values, time_units='ms')
```
gviejo, line 397: It's too tedious. I would just reshape df.values with 2 columns and call intervalSet. 
Also it's possible that some files contains a start time, a middle time and an end time. I don't think we can assume it's always start and end.

pynapple/io/neurosuite.py
```diff
@@ -372,6 +372,35 @@ def load_lfp(self, filename=None, channel=None, extension='.eeg', frequency=1250
...
+    
+        Returns
+        -------
+        IntervalSet  
+    
+        """
+        isets = self.load_nwb_intervals(name)
+        if str(type(isets)) !="<class 'pynapple.core.interval_set.IntervalSet'>":
+            print("reading intervals...")
+            evt_file = os.path.join(self.path, self.path.split('/')[-1] + "." + name + ".evt")
```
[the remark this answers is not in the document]
GRVite: You actually call it like that, rem_ep = data.read_neuroscope_intervals('rem')
And yes, that is the name of the file. I will detail this in the description.

pynapple/io/neurosuite.py
GRVite: Changed

pynapple/io/neurosuite.py
```diff
@@ -382,7 +382,10 @@ def read_neuroscope_intervals(self, name):
     
         Parameters
         ----------
-        name: name of the .evt file
+        name: name of the .evt file. The name of the file in the loading directory should
```
gviejo, line 385: The numpy docstring format should be 
name: str
description

pynapple/io/neurosuite.py
```diff
@@ -394,9 +397,8 @@ def read_neuroscope_intervals(self, name):
             print("reading intervals...")
```
gviejo, line 397: Hum maybe not necessary :upside_down_face:

gviejo: The print I mean :upside_down_face:

pynapple/io/neurosuite.py
```diff
@@ -394,9 +397,8 @@ def read_neuroscope_intervals(self, name):
...
             evt_file = os.path.join(self.path, self.path.split('/')[-1] + "." + name + ".evt")
```
gviejo, line 398: It should be self.basename and not dependant on the path

pynapple/io/neurosuite.py
```diff
@@ -394,9 +397,8 @@ def read_neuroscope_intervals(self, name):
...
             df = pd.read_csv(evt_file, delimiter=' ', usecols = [0], header = None)
-            start = nap.Ts(df.iloc[::2].values, time_units='ms')
-            end = nap.Ts(df.iloc[1::2].values, time_units='ms')
-            isets = nap.IntervalSet(start.index, end.index, time_units='ms')
+            isets = nap.IntervalSet(df.iloc[::2].values, 
+                        df.iloc[1::2].values, time_units='ms')
             print(isets)
```
gviejo, line 402: There is a print left.

pynapple/io/neurosuite.py
gviejo: I would do something like this : 
```
def read_neuroscope_intervals(self, pathtofile=None, name=None):
      if pathtofile is None and name is not None:
             pathtofile = os.path.join(self.path, self.basename + "." + name + ".evt")
     else:
            raise RuntimeError('Please specifiy a path or name")
```

GRVite: Alright, (;

PR #81, opened 2022-03-02
read neuroscope intervals modified

GRVite:


pynapple/io/neurosuite.py
```diff
@@ -371,7 +371,43 @@ def load_lfp(self, filename=None, channel=None, extension='.eeg', frequency=1250
...
+        path2file: str
+            Path of the file you want to load.
+    
+        Returns
+        -------
+        IntervalSet
+            Containing two columns corresponding to the start and end of the intervals.
+    
+        """
+        isets = self.load_nwb_intervals(name)
```
gviejo, line 398: I would add a condition ``` if name: ``` before calling load_nwb_intervals.

pynapple/io/neurosuite.py
```diff
@@ -371,7 +371,43 @@ def load_lfp(self, filename=None, channel=None, extension='.eeg', frequency=1250
...
+        if str(type(isets)) !="<class 'pynapple.core.interval_set.IntervalSet'>":
```
gviejo, line 399: I would remove this. if the previous condition is good no need to go through this.

gviejo (changes requested):

Issue #97, opened 2022-06-29
Flexible loading of KS data


vigji:
I am using the class to load Phy data, for spikes extracted with KS3. I am seeing that the loading is hardcoded to only take `"good"` spikes; I'm wondering if this could be changed to be flexible, giving the user the choice to also include eg `"mua"` if they want to for whatever reason.

I can see this happening in 2 ways (but I'm not familiar with the whole framework yet):
1. Passing an (optional) argument to to the `Phy` loader specifying the keys that are good to be loaded; this is straightforward but comes with the annoying layer of having to delete and regenerating the folder to switch eg from an analysis including everything to an analysis including only good spikes
2. Loading all files by default and add some internal filtering, that can be toggled during the analysis to expose only a fraction of the internally loaded spikes through the `self.spikes` attribute (eg changing it as a property).

Happy to work on this if you think might fit with the rest of the interface!

Cool work with the library! :)

vigji:
Happy to try a PR on it! Can you just give me pointers on where in the H5File would you like to keep the info on the label so that it can be reloaded with the rest?

gviejo:
Closing this issue as loading phy won't be part of pynapple anymore

Issue #96, opened 2022-06-29
Error loading Phy data


javierzsm:
Hi, 
I can't figure what's the issue here. I am trying to load some Phy data

My code:
```
import numpy as np
import pandas as pd
import pynapple as nap

data_directory = r'C:\Users\javier.zorrilla\Documents\NPX_OpenEphys_SampleData\kilosort3'

data = nap.load_session(data_directory, 'phy')

spikes = data.spikes
position = data.position
epochs = data.epochs

print(spikes)
```

and I get this error:

Traceback (most recent call last):

  File "C:\Users\javier.zorrilla\Anaconda3\envs\pynapple\lib\site-packages\spyder_kernels\py3compat.py", line 356, in compat_exec
    exec(code, globals, locals)

  File "\\l2export\iss02.bacci\trisomic-alpha5\javier\analysis\pynapple\crash_test_pynapple.py", line 10, in <module>
    data = nap.load_session(data_directory,

  File "C:\Users\javier.zorrilla\Anaconda3\envs\pynapple\lib\site-packages\pynapple\io\misc.py", line 53, in load_session
    return Phy(path)

  File "C:\Users\javier.zorrilla\Anaconda3\envs\pynapple\lib\site-packages\pynapple\io\phy.py", line 55, in __init__
    self.load_phy_spikes(path, self.time_support)

  File "C:\Users\javier.zorrilla\Anaconda3\envs\pynapple\lib\site-packages\pynapple\io\phy.py", line 134, in load_phy_spikes
    cluster_id_good = self.cluster_group[self.cluster_group.group=='good'].index.values

  File "C:\Users\javier.zorrilla\Anaconda3\envs\pynapple\lib\site-packages\pandas\core\generic.py", line 5575, in __getattr__
    return object.__getattribute__(self, name)

AttributeError: 'DataFrame' object has no attribute 'group'

vigji:
Getting same error here, after running KiloSort 3 on the data and creating the phy data folder (no manual curation with phy done on the data yet, though).
If I read the data frame directly with `pandas`, it seems that the column is not `group` but `KSLabel`, is this correct?

javierzsm:
@vigji thanks to your comment I found the issue. 

For some reason I was missing the cluster_info.tsv file.

I curated again the ks3 data and rewrote all the Phy output.

cluster_info.tsv is now properly loaded into a dataFrame and has a column named 'group'.

Thanks and sorry for the noise. I hope it helps somebody else.

[...]

PR #98, opened 2022-06-30
Group load

vigji:
Here it is! 
In this PR, I addressed:
 - #96 by making the column from which infos are loaded either "group" or "KSLabel" [here](https://github.com/iurillilab/pynapple/blob/c46d80b00abbc38a3782aa4a6069b8454a157817/pynapple/io/phy.py#L177)
 - In doing so, I refactored a bit the loading function, it had some brittle parts. I made `cluster_info`/`cluster_group` only `cluster_info`, and not making it an attribute as it is set and used only in the method to load from Phy folder.
 - There's also a minor fix in the rates calculation, for some reason with my very ugly data I have neurons with no spikes and was getting time supports of 0 duration; now frequency is set to np.nan in those cases. I think in general is a good fallback but happy to remove it if it is too specific for my edge case
 - I implemented what I suggested in #97: now spikes are all loaded, regardless of their classification, in the self._all_spikes attribute; and then selected (with caching, as the `TsGroup` indexing seems slow, 1-2s) when querying the self.spikes property, depending on what is specified in the `valid_group_labels` argument (by default `["good"]`). Labels are stored as a "label" unit attribute in the nab file. Back compatibility is ensued by the fact that if no label is found in the file, the filtering is just always returning all spikes.

Locally, tests are running. I don't find specifications for linting (are you considering implementing it? I'd recommend it!) so I don't know what to adhere to formatting-wise, feel free to let me know in case.
I will make sure I clean up prints and notebooks changes.
Let me know! And thank you for the package!

### DIFF
--- a/docs/notebooks/pynapple-io-notebook.ipynb
+++ b/docs/notebooks/pynapple-io-notebook.ipynb
@@ -3,7 +3,11 @@
   {
    "cell_type": "markdown",
    "id": "b582a340",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "# IO Tutorial\n",
     "\n",
@@ -21,7 +25,11 @@
    "cell_type": "code",
    "execution_count": 1,
    "id": "25fcec41",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
    "outputs": [],
    "source": [
     "import numpy as np\n",
@@ -33,7 +41,11 @@
   {
    "cell_type": "markdown",
    "id": "0f110346",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "In this [example dataset](https://www.dropbox.com/s/1kc0ulz7yudd9ru/A2929-200711.tar.gz?dl=1), the data contains a sample recording from the anterodorsal nucleus of the thalamus and the hippocampus, with both a sleep and a wake phase. It contains both head-direction cells (i.e. cells that fire for a particular direction of the head in the horizontal plane) and place cells (i.e. cells that fire for a particular position in the environment).\n",
     "\n",
@@ -51,7 +63,11 @@
    "cell_type": "code",
    "execution_count": 2,
    "id": "78ade0f2",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
    "outputs": [],
    "source": [
     "data_directory = '../../your/path/to/A2929-200711'\n",
@@ -62,7 +78,11 @@
   {
    "cell_type": "markdown",
    "id": "b571f99d",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "<img src=\"Screenshot1_baseloader.png\" width=700 height=700 />\n",
     "\n",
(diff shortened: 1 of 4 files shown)


vigji:
Also, tests are running locally on python 3.9. I still have to implement a specific test for this case, let me know if you're happy with me adding another file to the tests assets to deal with both possible scenarios of data with and without labels. 
Would you consider implementing GitHub Actions to automatise the rest of the testing in other python environments?

gviejo:
Hello Luigi, Thanks for the PR. I will test it as soon as possible next week.

gviejo:
Hey Luigi,
I wrote one comment about something I am not sure works. Maybe it's a problem of phy version.

I like the idea of caching the label good but I am not sure it provides enough flexibility. Looking at this issue, I am starting to think that the best is just to add all the metainfo available in the TsGroup and let the user decide after what to pick with groupby_category for example :
```
spikes = data.spikes.getby_category("KSLabel")['good']
```
or
```
mua = data.spikes.getby_category("UserLabel")["mua"]
```
So basically, the object data.spikes should have the following columns 
[(freq', KSLabel, UserLabel, shank, channel) (skipping them if it's not found in the cluster_info)


Ok for the rate fix otherwise. And for Github Actions, I need to look into it.

gviejo:
Hey Luigi, 

Here is the review I wrote :

image

But I think it doesn't matter now that we agree on returning both the KSLabel and the UserLabel for all units. 
Let me know if you want to make a new PR or if you want to update this one? I can also make the modifications myself.

And yes for the PR for github actions. If you can write somewhere a short description of what each file does, that would be great.
Thanks
Guillaume

vigji:
> But I think it doesn't matter now that we agree on returning both the KSLabel and the UserLabel for all units. Let me know if you want to make a new PR or if you want to update this one? I can also make the modifications myself.

All makes sense to me! I would keep it into this PR if it is fine for you, as there's already all changes related to loading those infos in the units file. So, should I just transfer all columns that can be find in the info file to new attributes of the units?

> And yes for the PR for github actions. If you can write somewhere a short description of what each file does, that would be great.

Great, than I would do another PR for that, I'll open an issue for the details!

gviejo:
Hello Luigi, what is the status of this PR? I think we agreed to load both the KSLabel and the UserLabel.

vigji:
I am happy to push this forward, but I need your feedback on the part that would change behavior of the code for current users. In the scenario we are envisioning `data.spikes` should return all available units I guess, but this will change the behavior of existing scripts that just use `data.spikes` with folders for which there is no `.nwb` file yet. Is this ok?

gviejo:
I need to check if there is a compartment in nwb to store and retrieve the KSLabel and UserLabel. Otherwise, I agree it would be hard to have the same information in data.spikes if you load from nwb versus you load from the phy files.

vigji:
No wait, that is not exactly what I meant: what I was raising is that, assuming that there is a way to store KSLabel and UserLabel in the nwb (it should be flexible enough to do it), then we have to deal with the default behavior of `data.spikes`. I would assume that it should return all spike stored, while `data.spikes.getby_category("UserLabel")["x"]` should filter spikes given some labels. But this means that all the existing code where `data.spikes` is used will change its behavior, this is what I am concerned about. The behavior won't change when working on a nwb file which was already generated filtering all "good" spikes, I agree, but still, maybe we want to raise a warn?

gviejo:
I think I see your point. And I think the warning solution is a good idea for already existing code.

gviejo:
Hey Luigi, I move the IO to [nwbmatic](https://github.com/pynapple-org/nwbmatic) as discussed in #130. Is it possible to close this PR and open a new one in nwbmatic if this is still an issue?

vigji:
Sorry, I thought I got back on this already! Sure, will test it out soon.

PR #105, opened 2022-08-10
linting and testing files

vigji:
This is the version with only the files required for the configuration. Ideally the changes should be applied in subsequent commits on this branch if that is fine with you, otherwise it is tricky to know whether testing is running! (GitHub Actions is already running on my fork, but failing until I apply the lint obviously). But you can also merge this and fix tests in a separate branch!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,54 @@
+name: tests
+
+on: push
+
+jobs:
+  lint:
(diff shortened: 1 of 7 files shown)


gviejo:
Thank you Luigi. I played around with github actions to lint and test. I will apply it to pynapple now. 
Also I will try to move the setup.py to pyproject.toml as suggested by the roadmap of PYPA for python packaging.

PR #116, opened 2022-11-30
Added Ts randomization routines

davidespalla:
Dear Guillaume and Pynapple maintainers,

I implemented some randomization procedures for pynapple ts and tsgroup objects, with the idea that they could be useful to compute null distributions for statistical testing.
I tried to follow the contribution guidelines as closely as possible. 
Here is a description of the new functionalities:

I located the routines in process.randomize.py. Currently, there are three types of randomization implemented:

- `shift_timestamps` shifts all the timestamps in a Ts object by a random amount, wrapping the end to the beginning. If applied to a TsGroup, all Ts in the group are independently shifted
- `jitter_timestamps` jitters all the timestamps in a Ts (or TsGroup) of a different random amount, uniformly sampled betw. -max_jitter and max_jitter. By default, the number of timestamps is preserved, and the timesupport can change, but the reverse can be enforced with keep_tsupport=True
- `resample_timestamps` re-draws the same number of timestamps in the time_support of the object.

The functions are implemented to handle a Ts or TsGroup objects, with an interface function exposed to the user that handles exceptions and then calls the relevant strategy depending on the input type.
Tests are currently covering mostly output consistency. I'll be happy to implement further tests if you think more coverage is needed.

I have in mind to implement randomization procedures for time series and time series frames as well. Before proceeding, I'll wait for your feedback. In particular:

- Is a submodule of nap.process the best place to fit these functionalities?
- How is best to handle the different data types? Would it be preferable to have an explicitly different function for each core data type (so split the ts and tsgroup functions, for example?)
- Is it better to have the randomize.py file only contain the functions exposed to the user, not to overcrowd the documentation? 

Thanks for reviewing my PR. I hope these functionalities can be useful.

### DIFF
--- a/docs/process.randomize.md
+++ b/docs/process.randomize.md
@@ -1,0 +1,11 @@
+::: pynapple.process.randomize
+	handler: python
+	selection:
+		docstring_style: numpy
+	rendering:
+		show_root_heading: false
+		show_source: true
+		show_category_heading: false
+		members_order: source
+		
+
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -25,6 +25,7 @@ nav:
       - Tuning curves: process.tuning_curves.md
       - Decoding: process.decoding.md
       - Peri-Stimulus: process.perievent.md
+      - Randomzation: process.randomize.md
   # - Plugins:
   #   - pynalog: pynalog.md
   #   - pynapplesonfire: pynapplesonfire.md
--- a/pynapple/process/__init__.py
+++ b/pynapple/process/__init__.py
@@ -2,3 +2,4 @@ from .correlograms import *
 from .decoding import *
 from .perievent import *
 from .tuning_curves import *
+from .randomize import *
--- a/pynapple/process/randomize.py
+++ b/pynapple/process/randomize.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+# @Author: dspalla
+# @Date:   2022-01-30 22:59:00
+# @Last Modified by:   dspalla
+# @Last Modified time: 2022-11-17 17:16:16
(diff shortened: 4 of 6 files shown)


gviejo:
Hey Davide, 
Thank you for the pull request. I haven't tested it in details but it looks great. For  your questions,
1. I think nap.process like you did is the best place
2. To avoid functions proliferation, I think it's best to have only one function that can accept multiple data types
3. It's a good point. I need to check mkdocs for the proper way to skip some functions in a file. So far let's consider that all functions in randomize can appear in the documentation.

Also, I would like to suggest a fourth type of randomization: to randomize by ISI. You compute the ISIs, shuffle their position and convert back to Ts. 
Let me know if it's possible to add it to your PR. Otherwise we can always add it later.
Thanks again for the work,

Guillaume

davidespalla:
Dear Guillaume,
thanks a lot for your feedback.
I'll add the ISI randomization soon, and then update the PR.
Thanks a lot for your time.

Davide

davidespalla:
Hi Guillaume,
I added the ISI randomization (I called it `shuffled_ts_intervals()`, to be agnostic to what the timestamps are, but it can be easily changed if another name is preferred).
Let me know what you think about it, I remain available for any changes.
Cheers,

Davide

gviejo:
Thank you Davide, it's amazing.
I was about to bump a new version of pynapple so I am adding your functions to it. 

If I may give one last suggestion, it is to make a notebook at some point to put in the pynapple/docs that explain the pros and cons of each shuffling methods. Some real examples of how to use them in this notebook will probably be useful to the community.

If you have any other ideas of features, please don't hesitate to contribute again.

PR #125, opened 2023-03-23
Fixing Threshold

dhruvm9:
The threshold function now can consider values greater than or equal to, as well as lesser than or equal to the threshold.


dhruvm9:
Same edits + fixed linting

PR #131, opened 2023-05-16
fix channel loading in load_phy_params()

slcalgin:


gviejo (commented):
Hi could you give a little detail about the changes?

slcalgin:
Hi Guillaume.

It seems that depending on the sorting algorithm/wrapper function used before Phy, the output of channel_maps and other supporting files varies. For example, using Kilosort's reztoPhy2 wrapper creates a compatible channel_maps.npy file, which is what the original Pynapple Phy loader was written for. However, some wrappers for Phy, such as that of SpyKingCircus, do not organize channel_maps by shank - instead, it produces an auxiliary file called channel_shanks.npy, which is where the shank information is stored. 

In the updates, I'm checking for the presence of the channel_shanks file. If it is there, I load the Phy params accordingly, which fixes the issues we discussed a few days ago. If there is no channel_shanks file (but there is a channel_maps file), the original code is run. 

Let me know if anything is not clear!

gviejo:
Hi Selen, 

The linting test was not successful. You need to make sure you run the black command in your terminal to check for linting. Here are the commands to run in your terminal to make sure the tests will pass:
```
    black --check pynapple
    isort --check pynapple
    flake8 pynapple --max-complexity 10
```

Also I am not sure about this part. It seems to me that if the file `channel_map.npy` is not found, the code will never look for the `channel_shank.npy` and jump straight to the error. Maybe it's a problem of indentation?

```python
        if os.path.isfile(os.path.join(path, "channel_map.npy")):

            channel_map = np.load(os.path.join(path, "channel_map.npy"))

            if os.path.isfile(os.path.join(path,"channel_shanks.npy")):
  
                  channel_shank = np.load(os.path.join(path,"channel_shanks.npy"))

```

slcalgin:
Hi Guillaume,

Sounds good, I will try the tests you provided and get back to you.

As for your concern, this is what it is meant to do. Whether or not channel_shanks.npy exists, the program _needs_ channel_maps.npy. 

The difference now is that if we detect that channel_shanks.npy exists, then we have to adjust the way we read the channel maps. If channel_shanks.npy doesn't exist, then the old code is run. 

This is because it seems as though some programs (such as Spyking Circus) will not store channel shank information directly within channel_maps.npy (which is what the original code assumes, which is why we were encountering those strange formatting errors a few weeks ago) and will also create channel_shanks.npy. However, we cannot have channel_shanks.npy without channel_maps.npy. If there is _no_ channel_shanks.npy, then we assume that all needed information is in channel_maps.npy.

gviejo:
Ok I understand now. If you commit the changes for the tests to pass, it should be good to merge.

slcalgin:
Done!

gviejo (approved):

Issue #130, opened 2023-05-08
Load data from KS without the GUI


GRVite:
Hello! Is there a method to create the .nwb file without utilizing the GUI? The GUI is useful, but not very efficient for loading multiple sessions. I am unsure if there is an existing solution for this issue. 
If it is necessary to implement a new approach, what would be the most effective way to accomplish this? How to take advantage of the existing code?

vigji:
I agree that this would be a very convenient feature!
Maybe the information to use to fill in the interface metadata could be parsed from a suitably formatted `.json` file if it can be found in the data folder?

gviejo:
Referencing  #134 #133 #132 #131 #98 #120 #110 #97 #95

Hi guys (and welcome back @vigji )

This is a general responses regarding all the issues with the pynapple.IO. I would like to detail here the plan of what is gonna happen in the coming weeks. You can see the general idea in the image below :
![New_pynapple_io](https://github.com/pynapple-org/pynapple/assets/4353798/6b59f570-1af4-43c7-bb0b-4f83bc8ba4c5)

1. First of all, you guys should check [neuroconv](https://neuroconv.readthedocs.io/en/main/index.html). It's basically the same idea as the current pynapple.IO but it supports many more formats. @vigji I would divert the effort for loading phy data to this package which is more advanced than the pynapple IO. It will also favors uniformity across NWB. I can still review and accept your recent PRs.
2. Consequently, the current pynapple.IO has become quite irrelevant now especially if the people at catalystneuro are actively maintaining neuroconv. The idea is to deprecate the current pynapple.IO and move the code to another standalone package [nwbmatic](https://github.com/pynapple-org/nwbmatic). People who want to use the GUI can still install it and use it for creating NWB file. It will also reduce dependencies for pynapple (e.g. tiff, zarr) so better for the long term stability. Ideally, nwbmatic should rely a lot on neuroconv to ensure uniformity here as well.
3. The new pynapple IO will then contain one class to interface with NWB in the most user friendly way (tbd). There is a lot to be done here but the transition from a NWB file to a pynapple object should be as smooth as possible.
[...]

PR #132, opened 2023-06-06
Allow multiple label names in the column for spikes filtering for KiloSort 3.0

vigji:
After a long delay (had to stop working on the data) - I'm back to the issues we were discussing in #96, #98, #97. 

In this PR I'm just fixing a small issue arising from a column name inconsistency in the `cluster_group.tsv` file for KiloSort 3.0.

Was there anything done for the issue described in #97 ? If not, happy to keep working on that feature!

### DIFF
--- a/pynapple/io/phy.py
+++ b/pynapple/io/phy.py
@@ -50,7 +50,6 @@ class Phy(BaseLoader):
             self.nwb_file = list(self.nwb_path.glob("*.nwb"))[0]
         except IndexError:
             self.nwb_file = None
-        print(self.nwb_file)
 
         # Need to check if nwb file exists and if data are there
         # if self.path is not None:  -> are there any cases where this is None?


vigji:
I've extended this PR a bit further. I realised some other small issues with the Phy class that I have addressed, hoping that this is not a problem - if it is, happy to revert to the small initial goal. All changes are feature-neutral; tests are passing, behavior of the class should be unaltered.

- I removed the path argument sent around to several methods; this value is already available in self.path
- I streamlined the flow a bit; there were in multiple places several nested conditions that could be flattened out/shortened for the sake of readability 
- In this context, I've moved the filesystem handling to the `pathlib` library, which is recommended since `python>3.4` as it makes for shorter and more readable patterns. If this is an issue, I can revert; if it is not, I'm happy to pursue the refactoring further!
- I've noticed that at some point there is the following pattern to load parameters:

  ```python
  sys.path.append(path)
  import params as params
  params.sample_rate
  ```
  I guess this is reading some parameters defined as variables in a  `.py` files, but it is a pretty convoluted syntax conflating code and parameters - I guess that a general solution for this (also going in the direction of #130 ) would be to support a parameter configuration stored in a `.txt`, a `.json` or `yml` file. I've not removed the code as I don't know if for some users this is important, but I'd recommend to discontinue it!
- I also noticed that there's a full method (load_lfp)  that seems duplicated from the `Neurosuite` class, and probably would not work in the `Phy` class as it is missing the load_neurosuite_xml method and the `nChannels` property. If this is the case, I would suggest to either fix the code or remove the method for now if unused! Left unchanged for now
- Some of the path handling stuff, like the search for preloaded data, is happening in both the parent class and the children, and could instead be kept just in the parent. I'm happy to fix this if I pursue the `os` -> `pathlib` transition, for now I left unchanged.

vigji:
(_Edit_) I actually noticed that the parameters are stored in the .py file from the guy and loading is actually necessary. This gives an error if the `params` library is installed in the environment! I will address this in a separate PR.

vigji:
(the test failure looks stochastic and the same seems to happen also in `main`)

PR #144, opened 2023-07-06
Adding Allen Brain Atlas loader

slcalgin:
New loader: io/allends.py: AllenDS
Currently supports Allen Neuropixels data
New package requirement: allensdk

Edit--more info about the code:

This loader utilizes the allensdk API to access the needed data. It loads epochs, various stimulus epochs, and spikes (along with unit metadata) into Pynapple core objects. Additional data that I thought would be useful is loaded in a dictionary of DFs under the attribute "metadata". Should the user want data that we didn't load from Allen, they (or we) can easily adjust the code to load it into metadata (since metadata loading doesn't restructure data from Allen). 

To actually access and download the data, the user provides a path to which they want the Allen data downloaded. Within this path, the loader creates a folder called "pynallensdk". This is the directory fed into the API. So cache data (i.e. data non-specific to a session) is downloaded here. These files include channels.csv, probes.csv, etc. Then the user is prompted to select a session ID from a GUI. Once the user does this, the session specific .nwb file is stored under session_#x/pynapplenwb. To get the file downloaded there, the loader makes a manifest.json file that specifies this prior to the data download. (When I added this capability, I was originally calling the super's functions that needed the .nwb file in the "pynapplenwb" folder. I have since changed the way the loader works, but for the sake of consistency with the rest of the Pynapple i/o system, I kept it.) 
If the user has previously worked with this same session (given the same original path), then the data won't be re-downloaded, since the Allen API is smart. Downloading the data the first time takes a little bit of time. 

One thing I would like to improve on is the loading time. The process of retrieving data from API and reformatting such a large amount of data makes the loader pretty slow. The .nwb file system we have currently won't solve this dramatically. I think in the future moving away from .nwb files and resorting to using .npy files might solve this issue after the first load. 

If anything isn't clear, let me know. It sounds a bit convoluted here but its pretty simple. 

Note: The Neuropixels API downloads data locally to user's computer. However, this loader never directly accesses the local data and instead accesses all data through the API. Therefore is Allen ever decides to give users the option not to download anything locally, the code should still be usable.


slcalgin:
Hi! I meant to add metadata_columns to another branch so I reverted it.

PR #177, opened 2023-09-22
Pynapple paper tutorials 

dhruvm9:
Peyrache et al 2015 and Zheng et al 2022 dataset figures have been converted to tutorials


dhruvm9:
Removed images folder

PR #189, opened 2023-10-15
Dev

gviejo:


pynapple/core/time_series.py
```diff
@@ -171,6 +171,18 @@ def shape(self):
...
+    def size(self):
+        return self.values.size
+
+    @property
+    def min(self):
+        return self.values.min()
+
+    @property
+    def max(self):
+        return self.values.max()
```
kushalkolar, line 184: Sometimes the user may want the min, max over an axis. It might be better to just wrap values.max() and pass any *args, **kwargs.

gviejo: Would it be ok for fastplotlib? From your example, it seemed that min and max were attributes?

kushalkolar: Yes it would for fastplotlib but not necessarily other situations if they want to use some other for doing some sort of computation. For example. if `d` is a dask array and they want to compute the min/max along an axis which has specific chunking.

kushalkolar: Ah, that might be a good workaround, I wonder if dask, etc. also have that (probably, since they try to be numpy-like). Maybe something you might want to mark as an experimental feature for now, or just document the behavior so users aren't confused.

gviejo: There is a [notebook](https://pynapple-org.github.io/pynapple/generated/gallery/tutorial_pynapple_numpy/) to document it. 

Just for me to understand, do you need to wrap any numpy array into a new class that has the min and max property when giving it to ImageWidget?

kushalkolar: Nope, if it's a regular numpy array it subsamples the array and then uses the min, max methods. If they are properties it uses those values directly.

https://github.com/fastplotlib/fastplotlib/blob/25385bede5d31657ee263f9920b440a27667f38c/fastplotlib/utils/functions.py#L151

gviejo: I see. So the last np.nanmin, np.nanmax should work on pynapple TsdTensor. So realistically, there is no need for the new properties right?

kushalkolar: Ah ok so if nanmax, nanmin are already defined I should just use those in fastplotlib! It's most robust anyways. I'll try this out and let you know (might be a few days, exam this week). 
The other properties, shape, ndim, size are still required.

PR #224, opened 2024-01-26
Dev

gviejo:


gviejo:
Adding spike trigger average for all dimensions.

PR #228, opened 2024-02-07
Non-numpy array support

BalzaniEdoardo:
In this PR I worked on the support for non-numpy array data. 

##  Motivation
Originally, jax arrays could be used to initialize a pynapple Ts object. However, this would eventually create unexpected behaviors. For instance,

```python
import pynapple as nap
import jax.numpy as jnp

tsd = nap.Tsd(t=jnp.array([1,2,3]), d=jnp.array([1,2,3]))
ep = nap.IntervalSet(1.5,2.5)
print(type(tsd.restrict(ep)))
```
would print `<class 'pynapple.core.time_series.Ts'>`, converting to Ts.

## Design Choices

1. **Convert to NumPy**: Pynapple relies on `numba` for efficiency, many array-like objects in deep learning frameworks (jax, torch...) cannot be handled by `numba` (which compiles for CPU). It is impossible to rely on `numba` and operate on` jax` arrays . For this reason I opted for converting array-like objects to numpy raising a UserWarning. 

2. **Suppress UserWarnings**: Added a configuration class that can be used to suppress these warnings. This is intended for `jax` users that can silently convert between numpy and jax arrays. 

3. **Testing**:  created a MockArray class for testing. Included tests for desired behavior.


BalzaniEdoardo:
I removed the dependency on jax that was creating issues with python 3.8

gviejo (approved):
It's great.

Issue #243, opened 2024-03-13
Improve API to filter `TsGroup`


alejoe91:
Hi @gviejo 

As mentioned previously, this is to suggest an easier API to filter `TsGroup`. Specifically, this is how filtering the units table from the `pynwb.Units` object looks like:

```
units = nwbfile.units[:]
# units is now a pandas dataframe
pass_qc = units[(units.isi_violations_ratio < 0.5) & 
                (units.firing_rate > 0.1) &
                (units.presence_ratio > 0.95) &
                (units.device_name == 'ProbeB')]
# or
pass_qc = units.query("isi_violations_ratio < 0.5 and ...")
```

The same operation in pynapple is much mroe convoluted:
```
spikes = data['units']

tsgroup = (spikes.getby_category('device_name')['ProbeB']
           .getby_threshold('isi_violations_ratio', 0.5, op='<')
           .getby_threshold('presence_ratio', 0.95, op='>')
           .getby_threshold('firing_rate',0.1,op='>'))
```

Would it be possible to simplify such queries with a simil-pandas API?

BalzaniEdoardo:
Hi Ale! Thanks for your suggestion. I have implemented a new `__getattr__` and a new `__getitem__` and `__setitem__` logic for the `TsGroup`. The PR is currently under revision, if you want to take a look. I'll ping you when it is merged!

Cheers
Edo

alejoe91:
> Hi Ale! Thanks for your suggestion. I have implemented a new `__getattr__` and a new `__getitem__` and `__setitem__` logic for the `TsGroup`. The PR is currently under revision, if you want to take a look. I'll ping you when it is merged!
> 
> Cheers Edo

Thannks Edo! I'll take a look and give it a try :)

gviejo:
Solved in #263

Issue #246, opened 2024-03-21
Make `TsGroup` a Pandas `DataFrame`?


jsiegle:
The `TsGroup` class behaves in many ways like a Pandas `DataFrame`. Given that a `DataFrame` is already being used under the hood to manage metadata for this class, it seems like making the whole thing into a `DataFrame` would be a natural choice.

Currently, there are a number of methods that are re-implemented in the `TsGroup` class that would come for free if it were a `DataFrame`:
- printing – Pandas has really nice printing methods, whereas the printout of a `TsGroup` is hard to read
- filtering – as mentioned in #243, the filtering syntax for a `TsGroup` is unnecessarily complex
- accessing columns – with a `DataFrame` I can simply call df.<column_name>, whereas `TsGroup` has reimplemented this as `get_info` (which ends up returning a Pandas `Series` object anyway)
- adding columns – adding a column to a `DataFrame` is done via bracket syntax, while `TsGroup` has a `set_info` method for this
- saving / loading – a `TsGroup` can only be saved as an .npz file and requires 4 lines of code to load, whereas Pandas supports a variety of file formats

There are of course many things (for example slicing or sorting) which can currently only be done with a `DataFrame`. And the unique methods of `TsGroup`, such as `restrict`, `value_from`, and `count`, would be straightforward to build on top of a `DataFrame`.

Therefore, my impression is that having `TsGroup` inherit from a `DataFrame` would reduce code complexity, make the class easier to interact with, and add substantial new functionality. Would the Pynapple devs consider doing this? If not, what are the reasons to avoid it?

BalzaniEdoardo:
Hi Josh, 

Thank you for your detailed suggestions regarding the `TsGroup `class. As mentioned in the PR #263, we addressed most of your concerns. Now with `TsGroup` you can:

- filter using boolean or integer array-like objects.
- access assess columns using the `tsgroup.column_name` syntax.
- Add metadata `tsgroup["metadata"] = mydata`

We are also planning to improve the `__repr__ `for better readability.

We discussed the possibility of having `TsGroup` inherit directly from `DataFrame`. However, our experience has shown that direct use of pandas can sometimes lead to accidental manipulations of the time-axis, which is critical to avoid in `pynapple`. The current architecture of `TsGroup `helps maintain strict control over time-axis manipulations, ensuring the reliability of the data.

Cheers,
Edoardo

gviejo:
[...]

PR #263, opened 2024-04-05
Ts group getitem

BalzaniEdoardo:
## Content
This PR solves issue #243 And also addresses most of the issues raised in #246.

#### **`__getattr__`  method added.**
Defined a  `__getattr__` that allows the following syntax,
  
```python
meta_data_series = tsgroup.name  # get the column "name" of `self._metadata` DataFrame
```
Raises an error if the name is not in the metadata.

#### **`__setitem__`  method modified.**
TsGroup is now mutable, but allowing only the metadata to change after initialization. The `__setitem__` allows only string as input, and enables,
```python
self._metadata[key] = values
```
Raises error if the key is non-string.

#### **`__getitem__`  method modified.**
The `__getitem__` of TsGroup  logic was modified as follows,

1. If one of the keys in `self.keys()` is passed, then the corresponding `Ts` is returned.
2. If the string is in the self._metadata.columns (hence it is a `str`, hence it is not in self.keys()), then it returns  self._metadata[key]
3. If it is a `bool` ArrayLike of dim 1 it returns a TsGroup including the Ts corresponding to `keys = np.where(bool_array)[0]`
4. If it is an `int`ArrayLike of dim 1 it returns a TsGroup including the Ts corresponding to the array entries.
5. Raises appropriate errors if none of the above conditions hold

### DIFF
--- a/pynapple/core/ts_group.py
+++ b/pynapple/core/ts_group.py
@@ -7,6 +7,7 @@
 import os
 import warnings
 from collections import UserDict
+from collections.abc import Hashable
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


alejoe91:
@BalzaniEdoardo seems to be working smoothly. Fixes #243

gviejo (approved):

gviejo:
Looks great. Last thing for TsGroup is to polish the `__repr__` function. I can work on that.

PR #300, opened 2024-06-07
Signal processing v0

kippfreud:
This PR contains FFT and Wavelet decomposition support for pynapple objects. Some functions are based on those from pywavelets package, but have, at this point, been changed significantly


gviejo:
I merged main into dev

gviejo:
Steps for linting are in the tox file.
https://github.com/pynapple-org/pynapple/blob/main/tox.ini

gviejo:
Note after quickly looking at the PR : pynapple uses numpy docstring not google docstring

kippfreud:
> Note after quickly looking at the PR : pynapple uses numpy docstring not google docstring

Docstings updated and linting fixed

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,224 @@
...
+        Sampling rate, in Hz.
+    freq : float
+        Center frequency of bandpass filter.
+    n_cycles : float, optional, default: 7
+        Length of the filter, as the number of cycles of the oscillation with specified frequency.
+    scaling : float, optional, default: 0.5
+        Scaling factor for the morlet wavelet.
+    wavelet_len : int, optional
+        Length of the wavelet. If defined, this overrides the freq and n_cycles inputs.
+    norm : {'sss', 'amp'}, optional
```
BalzaniEdoardo, line 191: is `sss` standard?

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    x = np.linspace(-scaling * 2 * np.pi, scaling * 2 * np.pi, M)
+    return np.exp(1j * ncycles * x) * (np.exp(-0.5 * (x**2)) * np.pi ** (-0.25))
+
+
+"""  
+The following code has been adapted from functions in the neurodsp package:  
+https://github.com/neurodsp-tools/neurodsp  
+
+..todo: reference licence in LICENCE directory
```
BalzaniEdoardo, line 53: docstrings should be numpydoc

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,224 @@
...
+
+import pynapple as nap
+
+
+def compute_fft(sig, fs):
+    """
+    Performs numpy fft on sig, returns output
+    ..todo: Make sig handle TsdFrame, TsdTensor
+
+    :param sig:    :param fs:    :return:
```
BalzaniEdoardo, line 18: numpydoc

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,224 @@
+"""
+    Signal processing tools for Pynapple
+
```
BalzaniEdoardo, line 3:
```suggestion
Signal processing tools for Pynapple.
```

BalzaniEdoardo (changes requested):
Great to see the code shaping up! I left a few notes, nothing major really

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,224 @@
...
+    Returns
+    -------
+    freqs: 1d array
+        Frequency indices.
+    """
+    return np.arange(freq_start, freq_stop + freq_step, freq_step)
+
+
+def compute_wavelet_transform(sig, fs, freqs, n_cycles=7, scaling=0.5, norm="amp"):
```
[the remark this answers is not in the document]
BalzaniEdoardo: adding to this, MATLAB has a nice discussion here about border effect and how to  think about them; It is a good ref also for the documentation of the method

https://www.mathworks.com/help/wavelet/ug/boundary-effects-and-the-cone-of-influence.html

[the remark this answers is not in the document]
kippfreud: 1. Yes this is true, I have fixed in newest commit
2. sig.convolve currently has the same default behaviour as numpy convolve, which is used in the scipy package. This creates border artefacts but maintains signal length, which feels most appropriate. Other modes could also be implemented into the convolve function to match numpy's functionality: https://numpy.org/doc/stable/reference/generated/numpy.convolve.html
3. I could and will in the next commit, however this decomposition is very memory intensive when using TsdFrames of any significant length, and TsdTensors would be even more so. Perhaps it is worth adding the functionality in but adding a warning?

pynapple/process/signal_processing.py
kippfreud: Both `sss` and `amp` are taken from the neurodsp package [compute_wavelet_transform](https://neurodsp-tools.github.io/neurodsp/generated/neurodsp.timefrequency.compute_wavelet_transform.html) function. They are not present in the scipy and PyWavelets compute_wavelet_transform functions, which do not have normalisation procedures built into them. 

In LFP literature, I have not seen normalisation like `sss` before on wavelets. I have seen `amp` be used, as well as something called `mad` normalisation be used. Should I remove `sss`?

pynapple/process/__init__.py
```diff
@@ -15,6 +15,7 @@
     shift_timestamps,
     shuffle_ts_intervals,
 )
+from .signal_processing import *
```
gviejo, line 18: I would specify which functions to import here

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+"""
+
+from itertools import repeat
+
+import numpy as np
+
+import pynapple as nap
+
+
+def compute_fft(sig, fs):
```
gviejo, line 13: I would call it `compute_spectrum` or `compute_periodogram`

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+    """
```
gviejo, line 14: Probably needs more description here. 
Also we need to explicit about the method. There should be a function call `compute_welch_spectrum` or `compute_welch_periodogram` somewhere. It can also be the same function but with different arguments for the methods.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Parameters
+    ----------
+    sig : pynapple.Tsd
+        The signal time series to obtain time frequency decomposition from
+    fs : int / float
```
gviejo, line 20: I would write `fs=None`. If None, compute `fs` by averaging the diff between two time indices. We need to write somewhere that this functions assume constant sampling rate.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+
+import pynapple as nap
+
+
+def compute_fft(sig, fs):
+    """
+
+    Parameters
+    ----------
+    sig : pynapple.Tsd
```
gviejo, line 18: Should work for any dimensions. 

`sig : Tsd, TsdFrame or TsdTensor`

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+        The signal time series to obtain time frequency decomposition from
+    fs : int / float
+        The sampling frequency of sig
+
+    Returns
+    -------
+    fft_result : np.ndarray
```
gviejo, line 25: A `pd.DataFrame` would be nice here.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+        fourier frequencies
+    """
+    if not isinstance(sig, nap.Tsd):
+        raise TypeError("Currently compute_fft is only implemented for Tsd")
+    fft_result = np.fft.fft(sig.values)
+    fft_freq = np.fft.fftfreq(len(sig.values), 1 / fs)
+    return fft_result, fft_freq
+
+
+def morlet(M, ncycles=5.0, scaling=1.0):
```
gviejo, line 37: I would call it `compute_morlet_wavelet`

gviejo: Also it would be nice in the notebook to show a plot of it.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+    M : int
+        Length of the wavelet
+    ncycles : float
+        number of wavelet cycles to use. Default is 5
+    scaling: float
+        Scaling factor. Default is 1.
+
+    Returns
+    -------
+    np.ndarray
```
gviejo, line 52: It could potentially be a Tsd object.

pynapple/process/signal_processing.py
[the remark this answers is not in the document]
gviejo: For convolve the limitation is number here. It only accepts mode same.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+
+def compute_wavelet_transform(sig, fs, freqs, n_cycles=7, scaling=0.5, norm="amp"):
+    """
+    Compute the time-frequency representation of a signal using morlet wavelets.
+
+    ..todo:: better normalization between channels
+
+    Parameters
+    ----------
+    sig : pynapple.Tsd or pynapple.TsdFrame
```
gviejo, line 131: It's usually `sig: Tsd or TsdFrame`

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+        Time series.
+    fs : float
```
gviejo, line 133: Add fs=None and try to infer it from the data

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,234 @@
...
+    if wavelet_len > sig.shape[-1]:
+        raise ValueError(
+            "The length of the wavelet is greater than the signal. Can not proceed."
+        )
+    morlet_f = morlet(wavelet_len, ncycles=n_cycles, scaling=scaling)
+    if norm == "sss":
+        morlet_f = morlet_f / np.sqrt(np.sum(np.abs(morlet_f) ** 2))
+    elif norm == "amp":
+        morlet_f = morlet_f / np.sum(np.abs(morlet_f))
+    mwt_real = sig.convolve(np.real(morlet_f))
```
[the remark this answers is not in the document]
kippfreud: You cannot; it will throw away the complex component.

`ComplexWarning: Casting complex values to real discards the imaginary part`

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,286 @@
...
+ax[1].set_title("Wake ephys")
+plt.show()
+
+
+# %%
+# There is much shared information between channels, and wake and sleep don't seem visibly different.
+# Let's take the Fourier transforms of one channel for both and see if differences are present
+channel = 1
+fig, ax = plt.subplots(1)
+fft_sig, fft_freqs = nap.compute_spectrum(sleep_minute[:, channel], fs=int(FS))
```
gviejo, line 124: This should be a pandas data frame so you can only return a single object. If you pass a TsdFrame, you can get a data frame where every column is one channel and the index are the `fft_freqs`.

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,286 @@
...
+        82.88,
+        117.19,
+        165.75,
+        234.38,
+        331.5,
+        468.75,
+        624.0,
+    ]
+)
+mwt_sleep = nap.compute_wavelet_transform(
```
gviejo, line 182: I would explain in more steps here.

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,286 @@
...
+freq = 3
+interval = (937, 939)
+wake_second = wake_minute.value_from(
+    wake_minute, nap.IntervalSet(interval[0], interval[1])
+)
+fig, ax = plt.subplots(1)
+ax.plot(wake_second.index.values, wake_second[:, channel], alpha=0.5, label="Wake Data")
+ax.plot(
+    wake_second.index.values,
+    mwt_wake.value_from(mwt_wake, nap.IntervalSet(interval[0], interval[1]))[
```
gviejo, line 211: `value_from` is not for restricting. You can use `restrict` or `get`

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,260 @@
...
+    fs : float, optional
+        Sampling rate, in Hz. If None, will be calculated from the given signal
+    """
+    if not isinstance(sig, nap.Tsd):
+        raise TypeError("Currently compute_fft is only implemented for Tsd")
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    fft_result = np.fft.fft(sig.values)
+    fft_freq = np.fft.fftfreq(len(sig.values), 1 / fs)
+    return fft_result, fft_freq
```
gviejo, line 33: This should return a pandas dataframe

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,260 @@
...
+    sig : pynapple.Tsd or pynapple.TsdFrame
+        Time series.
+    fs : float, optional
+        Sampling rate, in Hz. If None, will be calculated from the given signal
+    """
+    if not isinstance(sig, nap.Tsd):
+        raise TypeError("Currently compute_fft is only implemented for Tsd")
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    freqs, spectogram = welch(sig.values, fs=fs)
```
gviejo, line 51: In the notebook I would do a comparison of the difference between calling welch fft versus fft

gviejo (changes requested):

docs/examples/tutorial_signal_processing.py
kippfreud: I'm not sure what I could do without making private functions in signal_processing.py public. Perhaps I could make the Morelet wavelet function public, and plot some wavelets of different scales first? To give users an intuition of what a wavelet is?

pynapple/process/signal_processing.py
kippfreud: I will investigate further and add docs. These normalisation approaches both make the decomposition look a lot better, and PyWavelets (which most people who use wavelet will be familiar with) uses `sss` by default.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    fs : float, optional
+        Sampling rate, in Hz. If None, will be calculated from the given signal
+    """
+    if not isinstance(sig, (nap.Tsd, nap.TsdFrame)):
+        raise TypeError(
+            "Currently compute_spectogram is only implemented for Tsd or TsdFrame"
+        )
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    fft_result = np.fft.fft(sig.values, axis=0)
```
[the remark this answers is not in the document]
BalzaniEdoardo: look also the manual at pg 108. Another option would be return an FFT per epoch. unsure which would be the right output format for this. In any case, we should not treat the `sig.values` as a continuously sampled signal

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    fs : float, optional
+        Sampling rate, in Hz. If None, will be calculated from the given signal
+    """
+    if not isinstance(sig, (nap.Tsd, nap.TsdFrame)):
+        raise TypeError(
+            "Currently compute_welch_spectogram is only implemented for Tsd or TsdFrame"
+        )
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    freqs, spectogram = welch(sig.values, fs=fs, axis=0)
```
BalzaniEdoardo, line 54: same thing. we should figure out a way to treat separate epochs

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    fft_result = np.fft.fft(sig.values, axis=0)
+    fft_freq = np.fft.fftfreq(len(sig.values), 1 / fs)
+    return pd.DataFrame(fft_result, fft_freq)
+
+
+def compute_welch_spectogram(sig, fs=None):
+    """
+    Performs scipy Welch's decomposition on sig, returns output
+
+    ----------
```
BalzaniEdoardo, line 42:
```suggestion
    Parameters
    ----------
```

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    fft_result = np.fft.fft(sig.values, axis=0)
+    fft_freq = np.fft.fftfreq(len(sig.values), 1 / fs)
+    return pd.DataFrame(fft_result, fft_freq)
+
+
+def compute_welch_spectogram(sig, fs=None):
+    """
+    Performs scipy Welch's decomposition on sig, returns output
```
BalzaniEdoardo, line 40: Improve the description of the method. Short but understandable

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+
+    ----------
+    sig : pynapple.Tsd or pynapple.TsdFrame
+        Time series.
+    fs : float, optional
+        Sampling rate, in Hz. If None, will be calculated from the given signal
```
BalzaniEdoardo, line 46: add a "References" section to the welch method?

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    -------
+    np.ndarray
+        Morelet wavelet kernel
+    """
+    x = np.linspace(-precision, precision, M)
+    return (
+        ((np.pi * ncycles) ** (-0.25))
+        * np.exp(-(x**2) / ncycles)
+        * np.exp(1j * 2 * np.pi * scaling * x)
+    )
```
BalzaniEdoardo, line 83: If you store the morlet kernels in a (2, k) array (first axis is real/imagm, second axis is the number of kernels you have). you can call `nap.convolve` over it later on. `pynapple` will handle the convolution over both column dealing with the disjoint epochs.

this will have the additional benefit to work out of the box with pynajax

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+            )
+        n_cycles = iter(n_cycles)
+    return n_cycles
+
+
+def _create_freqs(freq_start, freq_stop, freq_step=1):
+    """
+    Creates an array of frequencies.
+
+    ..todo:: Implement log scaling
```
BalzaniEdoardo, line 123: does this render in mkdocs?

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+        Sampling rate, in Hz. If None, will be calculated from the given signal
+    """
+    if not isinstance(sig, (nap.Tsd, nap.TsdFrame)):
+        raise TypeError(
+            "Currently compute_welch_spectogram is only implemented for Tsd or TsdFrame"
+        )
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    freqs, spectogram = welch(sig.values, fs=fs, axis=0)
+    return pd.DataFrame(spectogram, freqs)
```
gviejo, line 55: Ok so wekch returns only the positive freqs so it's better to have the functions match indeed

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def _morlet(M=1024, ncycles=1.5, scaling=1.0, precision=8):
+    """
+    Defines the complex Morelet wavelet kernel
```
gviejo, line 60: Morlet

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+        output_shape = (sig.shape[0], len(freqs), *sig.shape[1:])
+        sig = sig.reshape((sig.shape[0], np.prod(sig.shape[1:])))
+    mwt = np.zeros(
+        [sig.values.shape[0], len(freqs), sig.values.shape[1]], dtype=complex
+    )
+    for channel_i in range(sig.values.shape[1]):
+        for ind, (freq, n_cycle) in enumerate(zip(freqs, n_cycles)):
+            mwt[:, ind, channel_i] = _convolve_wavelet(
+                sig[:, channel_i], fs, freq, n_cycle, scaling, norm=norm
+            )
```
BalzaniEdoardo, line 196: if the signal is a `TsdFrame`, you don't need to loop, you can call 

nap.convolve(sig, wavelet_kernel[0, i]), nap.convolve(sig, wavelet_kernel[1,i])

this handles the epochs and works with pynajax. the loop you'll need is on the filter bank.

I would argue that we should change `nap.convolve` to handle ND arrays so that the loop is done at the level of the convolve function instead, and here no loops are needed.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    if len(output_shape) == 2:
+        return nap.TsdFrame(
+            t=sig.index, d=mwt.reshape(output_shape), time_support=sig.time_support
+        )
+    return nap.TsdTensor(
+        t=sig.index, d=mwt.reshape(output_shape), time_support=sig.time_support
```
BalzaniEdoardo, line 202: the time support makes sense only if we do the wavelet convolution using the nap.convolve method which doesn't introduce artifact by treating the signal as equi-sampled

pynapple/process/signal_processing.py
gviejo: Also I noticed the two functions don't return the same scaled spectrogram

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    norm : {'sss', 'amp'}, optional
+        Normalization method:
+
+        * 'sss' - divide by the square root of the sum of squares
+        * 'amp' - divide by the sum of amplitudes
+
+    Returns
+    -------
+    array
+        Complex- valued time series.
```
BalzaniEdoardo, line 233: I would refactor this code to have:
1.  a function that creates the filter banks of all wavelets for the specified parameters and stores it in an NDArray.
2. a function that computes the normalization factor for different method
3. a function that calls the one that creates the filter bank, the normalization factor, uses `nap.convolve` over the filter bank and normalizes the output. 

The modularity of this approach means that if in the future want a new set of wavelets all you need to do is define a new filter bank creating function. On the wavelet class, you can now pass a simple string saying which wavelet you want to compute and it will call the appropriate filter bank generation.

BalzaniEdoardo (changes requested):
I left a few comment on how to improve this. We can discuss with @gviejo if you need clarifications.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    Notes
+    -----
+
+    * The real part of the returned array is the filtered signal.
+    * Taking np.abs() of output gives the analytic amplitude.
+    * Taking np.angle() of output gives the analytic phase.
+    """
+    if norm not in ["sss", "amp"]:
+        raise ValueError("Given `norm` must be `sss` or `amp`")
+    morlet_f = _morlet(int(2**precision), ncycles=n_cycles, scaling=scaling)
```
gviejo, line 244: `precision` does not appear in the docstring

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    # transform axis is always -1 due to the data reshape above
+    d = (coef.shape[-1] - sig.shape[-1]) / 2.0
+    if d > 0:
+        coef = coef[..., floor(d) : -ceil(d)]
+    elif d < 0:
+        raise ValueError(f"Selected scale of {scale} too small.")
+    return coef
+
+
+def _integrate(arr, step):
```
gviejo, line 264: The docstring is missing

pynapple/process/signal_processing.py
gviejo: I agree but that could potentialy lead to a very complicated way to do a tfd in pynapple

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+    if not isinstance(sig, (nap.Tsd, nap.TsdFrame)):
+        raise TypeError(
+            "Currently compute_spectogram is only implemented for Tsd or TsdFrame"
+        )
+    if fs is None:
+        fs = sig.index.shape[0] / (sig.index.max() - sig.index.min())
+    fft_result = np.fft.fft(sig.values, axis=0)
+    fft_freq = np.fft.fftfreq(len(sig.values), 1 / fs)
+    return pd.DataFrame(fft_result, fft_freq)
```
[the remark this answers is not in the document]
kippfreud: numpy fft also only returns only positive freqs

pynapple/process/signal_processing.py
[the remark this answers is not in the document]
kippfreud: After discussion with Guillaume, this function will only handle single epochs

tests/test_signal_processing.py
gviejo, line 1: I noticed there are a few error lines not covered. You can check it with the command `coverage report -m`. For example line 33.

gviejo: Also I noticed that none of the test actually compare values. For a particular input, you want to make sure that the output of a pynapple function matches what you would expect. For example for `compute_spectrogram` you can generate some fake data with a known frequency and see if the call of the function matches it with a call to `np.testing.assert_array_almost_equal` between the output and an array that you generates.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,273 @@
...
+    else:
+        output_shape = (sig.shape[0], len(freqs), *sig.shape[1:])
+        sig = sig.reshape((sig.shape[0], np.prod(sig.shape[1:])))
+
+    mwt = np.zeros(
+        [sig.values.shape[0], len(freqs), sig.values.shape[1]], dtype=complex
+    )
+
+    filter_bank = _generate_morlet_filterbank(freqs, fs, n_cycles, scaling, precision)
+    for f_i, filter in enumerate(filter_bank):
```
gviejo, line 194: From here, if `filter_bank` is an array, you don't need to do all the reshaping. `sig.convolve(filter_bank, ep)` should work directly.

gviejo (changes requested):

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,273 @@
...
+        Scaling factor.
+    precision: int.
+        Precision of wavelet to use.
+
+    Returns
+    -------
+    filter_bank : list[np.ndarray]
+        list of morlet wavelet filters of the frequencies given
+    """
+    filter_bank = []
```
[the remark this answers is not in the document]
kippfreud: Currently not, or with zero padding. I'm unsure why, some kind of border problem it seems.. I will discuss this with you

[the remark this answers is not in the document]
kippfreud: I have figured out how to do it, but if they were TsdFrame then what would the index be?

pynapple/process/signal_processing.py
kippfreud: It wont currently as the filter bank will be num_frequencies x length_of_frequency x 2(real and imag component) 
and convolve only works with 1 or 2 dimensional data

tests/test_signal_processing.py
kippfreud: This is now done by generating a 50hz sinusoidal, computing the cwt, then checking that the 50 hz frequency band is most powerful in the wavelet decomp

pynapple/process/signal_processing.py
kippfreud: time x wavepow in filter bank - conv once with real then with imag, then add

pynapple/process/signal_processing.py
kippfreud: plot will be added in API notebook

docs/examples/tutorial_signal_processing.py
```diff
@@ -244,325 +202,209 @@ def plot_timefrequency(
...
+# define our threshold
+threshold = 100
+# smooth our wavelet power
+window_size = 51
+window = np.ones(window_size) / window_size
+smoother_swr_power = np.convolve(
+    np.abs(mwt_RUN[:, ripple_freq_idx].values), window, mode="same"
+)
+# isolate our ripple periods
+is_ripple = smoother_swr_power > threshold
```
gviejo, line 337:
```suggestion
is_ripple = smoother_swr_power.threshold(threshold)
```

gviejo: Check the pynacollada notebook for this

gviejo (changes requested):

docs/examples/tutorial_wavelet_api.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+title = "Morlet Wavelet Filter Bank (Real Components): n_cycles=1.5, scaling=1.0"
+plot_filterbank(filter_bank, freqs, title)
+
+# %%
+# ***
+# Decomposing the Dummy Signal
+# ------------------
+# Here we will use the `compute_wavelet_transform' function to decompose our signal using the filter bank shown
```
gviejo, line 104:
```suggestion
# Here we will use the `compute_wavelet_transform` function to decompose our signal using the filter bank shown
```

docs/examples/tutorial_wavelet_api.py
```diff
@@ -0,0 +1,395 @@
...
+[ax[i].set_ylabel("Signal") for i in range(2)]
+plt.show()
+
+# %%
+# ***
+# Adding ALL the Oscillations!
+# ------------------
+# Let's now add together the real components of all frequency bands to recreate a version of the original signal.
+
+combined_oscillations = mwt.values.real.sum(axis=1)
```
gviejo, line 227:
```suggestion
combined_oscillations = mwt.sum(axis=1)
```

docs/examples/tutorial_wavelet_api.py
```diff
@@ -0,0 +1,395 @@
...
+ax.set_ylim(-6, 6)
+ax.margins(0)
+ax.legend()
+plt.show()
+
+
+# %%
+# ***
+# Parametrization
+# ------------------
```
gviejo, line 247: Looking at the plot, I can't see the difference between n_cycle=1, scaling = 1 vs n_cycle=3,scaling=2.0.

Either the parameters needs to be more extreme to actually make sense of the difference, otherwise it's not the right pair of parameters

docs/examples/tutorial_wavelet_api.py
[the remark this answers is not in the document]
gviejo: I think it's good but overall the difference between `n_cycles` and `scaling` should be more visually striking. What happens for `n_cycles` vs `precision`?

gviejo (changes requested):

docs/examples/tutorial_wavelet_api.py
kippfreud: This was a problem with the scaling of the wavelets - fixed in newest commit

docs/examples/tutorial_wavelet_api.py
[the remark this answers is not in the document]
kippfreud: I think this is fixed now, let me know what you think

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,268 @@
...
+        return nap.TsdFrame(
+            t=sig.index, d=cwt.reshape(output_shape), time_support=sig.time_support
+        )
+
+    return nap.TsdTensor(
+        t=sig.index, d=cwt.reshape(output_shape), time_support=sig.time_support
+    )
+
+
+def generate_morlet_filterbank(freqs, fs, n_cycles=1.5, scaling=1.0, precision=10):
```
gviejo, line 191: The filters still look stepped when zooming.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,268 @@
...
+        Length of the filter, as the number of cycles for each frequency.
+        If 1d array, this defines n_cycles for each frequency.
+    scaling : float
+        Scaling factor.
+    precision: int.
+        Precision of wavelet to use. Default is 8
+    norm : {None, 'sss', 'amp'}, optional
+        Normalization method:
+        * None - no normalization
+        * 'sss' - divide by the square root of the sum of squares
```
gviejo, line 135: Add more description here. The sum of squares along which dimension?

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,268 @@
...
+    convolved = convolved_real.values + convolved_imag.values * 1j
+    coef = -np.diff(convolved, axis=0)
+    if norm == "sss":
+        coef *= -np.sqrt(scaling) / (freqs / fs)
+    elif norm == "amp":
+        coef *= -scaling / (freqs / fs)
+    coef = np.insert(
+        coef, 1, coef[0, :], axis=0
+    )  # slightly hacky line, necessary to make output correct shape
+    cwt = np.swapaxes(coef, 1, 2)
```
gviejo, line 179: Ok I don't understand this. Maybe we can discuss it at some point.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Returns
+    -------
+    pynapple.TsdFrame or pynapple.TsdTensor : 2d array
+        Time frequency representation of the input signal.
+
+    Notes
+    -----
+    This computes the continuous wavelet transform at specified frequencies across time.
+    """
```
gviejo, line 146: Add a small example of how to call the function here

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,397 @@
...
+axd["ephys"].plot(
+    RUN_Tsd[:, 0].restrict(
+        nap.IntervalSet(
+            data["forward_ep"]["start"][run_index], data["forward_ep"]["end"][run_index]
+        )
+    ),
+    label="Traversal LFP Data",
+    color="green",
+)
+axd["ephys"].plot(
```
gviejo, line 106: No need for that. It doens't bring much information. Plus it's very artificial when the run stops and starts. Plot the LFP with a single color. The bottom plot help to show when the animal stops running.

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,397 @@
...
+    alpha=0.5,
+    label="LFP Frequency Power",
+    c="blue",
+    linewidth=2,
+)
+ax.set_xlabel("Freq (Hz)")
+ax.set_ylabel("Frequency Power")
+ax.set_title("LFP Fourier Decomposition")
+ax.set_xlim(1, 30)
+ax.axvline(9.36, c="orange", label="9.36Hz", alpha=0.5)
```
gviejo, line 156: No need for this. Eventually i would do a axvspan from 8 to 12 if you want to emphasize theta.

gviejo: Have you tried loglog or semilogy?

gviejo: Some example are shown here : https://www.nature.com/articles/s41593-020-00744-x

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,397 @@
...
+# ***
+# Isolating Ripple Times
+# -----------------------------------
+# We can see one significant peak in the 200Hz frequency power. Let's smooth the power curve and threshold
+# to try to isolate this event.
+
+# Define threshold
+threshold = 6000
+# Smooth wavelet power TsdFrame at the SWR frequency
+smoother_swr_power = (
```
gviejo, line 334: I would do step by step with more plots here.

docs/examples/tutorial_phase_preferences.py
```diff
@@ -0,0 +1,355 @@
+# -*- coding: utf-8 -*-
+"""
+Computing Phase Preferences
+============
+
+In the previous [Grosmark & Buzsáki (2016)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4919122/) tutorial,
```
gviejo, line 6: No need to reference the otehr tutorial

gviejo (changes requested):

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,268 @@
...
+        Step size of vgiven wave function array
+
+    Returns
+    -------
+    array
+        Complex-valued integrated wavelet
+
+    """
+    integral = np.cumsum(arr)
+    integral *= step
```
BalzaniEdoardo, line 267: I don't know how much precision matters here, but you can use the simpson method to get a better approximation of the numerical integral. may be slow to do it in a cumulative fashion but something to consider


```python
from scipy.integrate import simpson
import numpy as np
a,b = 0,1
func = lambda x: np.sin(x)
integr = lambda x: -np.cos(x)

xs = np.linspace(a,b,20)
print(f"true integral: {integr(b) - integr(a)}")
print(f"simps integral: {simpson(func(xs), x=xs)}")
print(f"euler integral: {np.sum(func(xs))*(xs[1] - xs[0])}")

# true integral: 0.45969769413186023
# simps integral: 0.4596975270737199
# euler integral: 0.48173554578218836
```

tests/test_signal_processing.py
```diff
@@ -0,0 +1,114 @@
+"""Tests of `signal_processing` for pynapple"""
+
+import numpy as np
+import pandas as pd
+import pytest
+
+import pynapple as nap
```
BalzaniEdoardo, line 7: If we are trying to match any other implementation (like pywavelet or similar) we should have a test that checks that the output of the functions we are matching are actually the same.

tests/test_signal_processing.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def test_compute_spectogram():
```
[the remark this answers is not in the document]
BalzaniEdoardo: try with a pure sinusoid, and make sure that the only peak shows up in the right location, try a weighted sum of 2 sinusoids too.
you can have as input `(freqs, amlplitudes)` as vector, then create a signal that is `signal = np.sum(amlplitudes * cos(freqs * time)))`, and check that the peaks are at the freqs, and the amplitudes are correct.

docs/examples/tutorial_signal_processing.py
kippfreud: I have but they did not look good. I will show you in our next meeting.

pynapple/process/signal_processing.py
kippfreud: The precision parameter controls the amount of time points at which the Morlet wavelet is evaluated at. One Morlet wavelet is created with 2**precision time points, and then it is stretched to calculate wavelet coefficients at different frequencies. This is why, at low frequencies (where the wavelet must be stretched a lot), with a low frequency you can see the wavelet repeating values.

tests/test_signal_processing.py
kippfreud: Because of some changes in logic e.g. using different convolve functions, the pywt wavelet decomposition and ours do not function exactly the same any more, and will not produce identical output.

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,361 @@
...
+ax0.set_title(f"EEG (1250 Hz)")
+ax1 = plt.subplot(gs[1,0])
+ax1.legend()
+ax1.plot(ripple_power.restrict(zoom_ep), label="150-250 Hz")
+ax1.plot(smoothed_ripple_power.restrict(zoom_ep))
+for s,e in rip_ep.intersect(zoom_ep).values:
+    ax1.axvspan(s, e, color='red', alpha=0.1, ec=None)
+ax1.legend()
+ax1.set_ylabel("Mean Amplitude")
+ax1.set_xlabel("Time (s)")
```
BalzaniEdoardo, line 359: zoom in to show the ripple

docs/api_guide/tutorial_pynapple_wavelets.py
```diff
@@ -0,0 +1,489 @@
+# -*- coding: utf-8 -*-
+"""
+Wavelet Transform
+=================
+
+This tutorial covers the use of `nap.compute_wavelet_transform` to do continuous wavelet transform. By default, pynapple uses Morlet wavelets.
```
BalzaniEdoardo, line 6: Add that wavelet are a great tool to capture changes of spectral characteristics of a signal over time. basically, why it is worth using wavelet transforms

docs/api_guide/tutorial_pynapple_wavelets.py
```diff
@@ -0,0 +1,489 @@
...
+    for a in [ax0, ax1]
+]
+
+
+# %%
+# ***
+# Adding ALL the Oscillations!
+# ------------------
+# Let's now add together the real components of all frequency bands to recreate a version of the original signal.
+
```
BalzaniEdoardo, line 226: maybe it is worth anticipating what is going to happen . We will now learn how to interpret the parameters of the wavelet, and in particular how to trade off the accuracy in the frequency decomposition with the accuracy in the time domain reconstruction; 

basically introducing the concepts before showing the initial failure. I think it makes for a better flow

BalzaniEdoardo (commented):
i left a bunch of stuff, mostly minor

pynapple/io/folder.py
```diff
@@ -4,7 +4,7 @@
 # @Author: Guillaume Viejo
 # @Date:   2023-05-15 15:32:24
 # @Last Modified by:   Guillaume Viejo
-# @Last Modified time: 2023-08-06 17:37:23
+# @Last Modified time: 2024-08-02 11:35:10
```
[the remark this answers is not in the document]
kippfreud: I don't believe we need this, @gviejo ?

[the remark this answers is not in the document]
gviejo: No you can remove the whole stuff

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,401 @@
...
+        raise ValueError("norm parameter must be 'l1', 'l2', or None.")
+
+    if fs is None:
+        fs = sig.rate
+
+    if sig.ndim == 1:
+        output_shape = (sig.shape[0], len(freqs))
+    else:
+        output_shape = (sig.shape[0], len(freqs), *sig.shape[1:])
+        sig = np.reshape(sig, (sig.shape[0], np.prod(sig.shape[1:])))
```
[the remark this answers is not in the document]
kippfreud: I do not believe these lines are equivalent when sig.ndim==1

This change causes the tests to fail.

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,401 @@
...
+    else:
+        raise TypeError("precision must be a float or int instance.")
+
+    filter_bank = []
+    cutoff = 8
+    morlet_f = _morlet(
+        int(2**precision), gaussian_width=gaussian_width, window_length=window_length
+    )
+    x = np.linspace(-cutoff, cutoff, int(2**precision))
+    int_psi = np.conj(morlet_f)
```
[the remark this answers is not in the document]
kippfreud: changing it to "morlet_f"

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,401 @@
...
+        Defines width of Gaussian to be used in wavelet creation.
+    window_length : float
+        The length of window to be used for wavelet creation.
+    precision: int.
+        Precision of wavelet to use. Defines the number of timepoints to evaluate the Morlet wavelet at.
+
+    Returns
+    -------
+    filter_bank : pynapple.TsdFrame
+        list of Morlet wavelet filters of the frequencies given
```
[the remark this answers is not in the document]
kippfreud: Agreed

tests/test_signal_processing.py
```diff
@@ -0,0 +1,533 @@
...
+        window_length = 1.0
+        fz = power.index
+        factor = np.pi**0.25 * gaussian_width**0.25
+        morlet_ft = factor * np.exp(
+            -np.pi**2 * gaussian_width * (window_length * (fz - f) / f) ** 2
+        )
+        assert np.isclose(
+            power.iloc[:, i] / np.max(power.iloc[:, i]),
+            morlet_ft / np.max(morlet_ft),
+            atol=0.1,
```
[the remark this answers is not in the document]
kippfreud: Agreed that it is a lax tolerance, this is due to the very minor difference between expected fft output and true fft output (there is some kind of shift which I do not understand). 

fft's are normalized so o.o.m is 1. We should discuss this difference tomorrow and see if we can identify the issue.

tests/test_signal_processing.py
```diff
@@ -0,0 +1,533 @@
...
+
+def test_generate_morlet_filterbank():
+    fs = 1000
+    freqs = np.linspace(10, 100, 10)
+    fb = nap.generate_morlet_filterbank(
+        freqs, fs, gaussian_width=1.5, window_length=1.0, precision=16
+    )
+    power = np.abs(nap.compute_power_spectral_density(fb))
+    for i, f in enumerate(freqs):
+        assert power.iloc[:, i].argmax() == np.abs(power.index - f).argmin()
```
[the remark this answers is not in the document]
kippfreud: comment added for now

docs/examples/tutorial_signal_processing.py
```diff
@@ -0,0 +1,361 @@
...
+ax.set_title("LFP Fourier Decomposition")
+ax.legend()
+
+
+# %%
+# The red area outlines the theta rhythm (6-10 Hz) which is proeminent in hippocampal LFP. 
+# Hippocampal theta rhythm appears mostly when the animal is running. 
+# (See Hasselmo, M. E., & Stern, C. E. (2014). Theta rhythm and the encoding and retrieval of space and time. Neuroimage, 85, 656-666.)
+# We can check it here by separating `wake_ep` into `run_ep` and `rest_ep`.
+run_ep = data['position'].dropna().find_support(1)
```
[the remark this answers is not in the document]
kippfreud: `# The run epoch is the portion of the data for which we have position data
run_ep = data['position'].dropna().find_support(1)
# The rest epoch is the data at all points where we do not have position data
rest_ep = wake_ep.set_diff(run_ep)`

pynapple/process/signal_processing.py
```diff
@@ -0,0 +1,401 @@
...
+
+    ret = pd.DataFrame(fft_result, fft_freq)
+    ret.sort_index(inplace=True)
+
+    if not full_range:
+        return ret.loc[ret.index >= 0]
+    return ret
+
+
+def compute_mean_power_spectral_density(
```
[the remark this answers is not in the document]
kippfreud: Later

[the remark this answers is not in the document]
kippfreud: @gviejo

tests/test_signal_processing.py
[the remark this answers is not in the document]
kippfreud: fixed

gviejo:
Great job. Thank you Kipp.

PR #305, opened 2024-06-26
Create a TsGroup from an iterable of Ts/Tsd objects

eschombu:
It doesn't seem to make a ton of sense to me to require that users manually package their timestamp/series objects into a dict when creating `TsGroup` objects, since the keys are required to be unique integers. I would guess that the vast majority of the time, users would be just doing `{0: ts0, 1: ts1, ...}`. Sure, this can be easily done by saying `data = dict(enumerate(ts_list))`, but why not just allow the user to say `data = ts_list` and do the dict creation automatically?

If the change described above is not desired, the second commit is probably still worth including. It ensures that the keys in `.data` and `.index` are in the same order by doing this:
```
data = {k: data[k] for k in self.index}
```
This might not be assured if `bypass_check == True`.

I added some tests for the new functionality. Again, any and all feedback/concerns/thoughts are welcome!

### DIFF
--- a/pynapple/core/ts_group.py
+++ b/pynapple/core/ts_group.py
@@ -138,6 +138,7 @@ class TsGroup(UserDict):
 
         data = {keys[j]: data[k] for j, k in enumerate(data.keys())}
         self.index = np.sort(keys)
+        data = {k: data[k] for k in self.index}  # Make sure data dict and index are ordered the same
 
         self._metadata = pd.DataFrame(index=self.index, columns=["rate"], dtype="float")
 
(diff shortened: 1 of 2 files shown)


gviejo (commented):
Looks good.

gviejo:
You can check the linting with the following commands :

    black --check pynapple
    isort --check pynapple --profile black
    flake8 pynapple --max-complexity 10
    coverage run --source=pynapple --branch -m pytest tests/
    coverage report -m

eschombu:
Reformatted with black, hopefully all good now. Let me know if you have any further comments/requests/suggestions.

gviejo (approved):

gviejo:
Great thanks Eric

PR #314, opened 2024-07-15
Option to have a Tsd as an index

vigji:
When working with Tsds, I might sometimes write, eg for filtering:

```python
some_tsd = Tsd(...)
some_tsd.d[some_tsd.d > 500] = np.nan
```

I think it would be nice and not confusing for how the class works atm (very thorough feature equivalence with np.ndarray) having instead:

```python
some_tsd = Tsd(...)
some_tsd[some_tsd > 500] = np.nan
```

Just by changing the setter and getter for the `Tsd`. 

Would something like this make sense? This is just a draft for `Tsd`, it could be extended to other classes maybe although I never encountered the need.


pynapple/core/time_series.py
```diff
@@ -104,6 +104,8 @@ def __init__(self, t, d, time_units="s", time_support=None, load_array=True):
 
     def __setitem__(self, key, value):
         """setter for time series"""
+        if isinstance(key, Tsd):
```
gviejo, line 107: I
```suggestion
        if isinstance(key, BaseTsd):
```

gviejo (changes requested):
i think there is a conflit with the previous PR

gviejo:
I think it's good and I would keep it only for time series.

pynapple/core/time_series.py
```diff
@@ -1237,14 +1239,18 @@ def __repr__(self):
                 return tabulate([], headers=headers) + "\n" + bottom
 
     def __getitem__(self, key, *args, **kwargs):
+        if isinstance(key, Tsd):
```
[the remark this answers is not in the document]
vigji: It is fixed now!

vigji:
Everything should have been addressed now, let me know if there's anything else I should fix

gviejo:
That's great. I would still add one test for it.

vigji:
Excellent point, adding tests I actually realised I did not set the behavior for `setitem` in `TsdFrame`. 

This should be now be good to go!

gviejo (dismissed):

gviejo:
It's missing for TsdTensor I think?

gviejo (dismissed):
On line 732

        if isinstance(key, BaseTsd):
            key = key.d

vigji:
Oh sorry maybe I misunderstood one of your early messages - do you thing this should be implemented for Tsdtensors as well?

vigji:
Or you mean I should exclude tensors?

gviejo:
Yes I would implement it for TsdTensor too

vigji:
I'll come back to this soon if still unaddressed!

Since I am at it, I'll ask a slightly related but separate question: why is `TsIndex` not a `Ts` but a `np.array`? Is there some actual reason or it is just historical?

gviejo:
Basically TsIndex is like pd.Index in pandas. `Ts` is the user facing module. `TsIndex` is the non-user facing module which hold the timestamps and timestamps methods across all modules with time.

vigji:
This should now good to go!

I did a minor refactoring of test classes for readability, hopefully that's fine. Also, I'd love to do a separate PR for tests cleanup if that is good for you!

gviejo (changes requested):
Looks good a few changes only

pynapple/core/time_series.py
```diff
@@ -772,15 +774,36 @@ def create_str(array):
...
-            index = self.index.__getitem__(key[0])
+        print("key", key)
+        if isinstance(key, Tsd):
+            if not np.issubdtype(key.dtype, np.bool_):
+                raise ValueError(
+                    "When indexing with a Tsd, it must contain boolean values"
+                )
+            output = self.values[key.values]
+            index = self.index[key.values]
+        elif isinstance(key, tuple):
```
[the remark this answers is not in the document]
vigji: if you are slicing over multiple dims at once. Maybe we do not want to deal with that?

pynapple/core/time_series.py
```diff
@@ -1205,15 +1231,18 @@ def __repr__(self):
                 return tabulate([], headers=headers) + "\n" + bottom
 
     def __getitem__(self, key, *args, **kwargs):
```
BalzaniEdoardo, line 1233: I think we can simplify the getitem by getting rid of the args and kwargs from the function signature:

1. removing the columns parameter passed to the getitem by `_TsdFrameSliceHelper`, which is overwritten anyways in the `__getitem__` of the TsdFrame.
2. removing the **kwargs passed to `TsdFrame` initialization in the  `__getitem__`  of TsdTensor. I don't see when a non-empty kwarg should passed to the TsdTensor `__getitem__` .

the test passes with this modification.

BalzaniEdoardo (commented):
Thanks Luigi! I did not go through the tests yet, the main code looks good to me, but I added a couple of suggestions;

pynapple/core/time_series.py
vigji: ok, makes sense!

vigji: done

vigji:
Let mw know if anything else is needed here! 

(In the test refactoring, move test generated files to temporary folders would be very useful to avoid spurious commits)

gviejo:
Thanks I will look into it as soon as possible.

gviejo (changes requested):
It's almost good. Is it possible to add this for all time series :

```

            if not all(
                np.issubdtype(k.dtype, np.bool_) if isinstance(k, Tsd) else True
                for k in key
            ):
                raise ValueError(
                    "When indexing with a Tsd, it must contain boolean values"
                )
```

I should only be boolean indexing with another Tsd otherwise it doesn't make sense to index with integers because it means you are breaking the relationship between the time index and the corresponding values. Somehow it's similar to shuffling but it should be a proper method in the shuffling module.

vigji:
Ok, done it! I also change a bit the logic to bit more interpretable, and fixed the tests accordingly

vigji:
Should have fixed stuff and rebased on dev head. Would be great if we can merge so that I won't have to resolve more conflicts in the future!

gviejo (approved):

gviejo:
Looks good thank you Luigi

PR #315, opened 2024-07-15
Pickling tsgroup

BalzaniEdoardo:
## Description

This PR solves the issue of infinite recursion for the pickling of "TsGroup".

Raises an error if `__getattr__` is called by the `__reduce__` and `__reduce_ex__` methods. 

These are attributes of the `pandas.Index`, which are the column names of the metadata, that would be called by captured by the `__getattr__` implementation otherwise, generating an infinite recursion.

## New Capability
This two line now do not generate an infinite recursion.

```python
import pickle
#...assume tsgroup is a TsGroup
pickled_tsgroup = pickle.dumps(tsgroup)
unpickled_tsgroup = pickle.loads(pickled_tsgroup)
```

### DIFF
--- a/pynapple/core/ts_group.py
+++ b/pynapple/core/ts_group.py
@@ -202,6 +202,10 @@ class TsGroup(UserDict):
         AttributeError
             If the requested attribute is not a metadata column.
         """
+        # avoid infinite recursion when pickling due to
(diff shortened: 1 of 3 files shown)


gviejo (approved):

gviejo:
Favoloso!

PR #322, opened 2024-07-30
309 get slice

BalzaniEdoardo:
# `get_slice` method for `Base`

This PR adds a `get_slice` method to `Base`, and will be inherited by `Ts`, `Tsd`, `TsdFrame`, `TsdTensor`.

The method returns a `slice` object that can be used to get a single time point of the Ts, or a range of time points.

It can operate in 3 modes:

1. "closest": the default mode, starts/ends the slice with indices closest to the start/end time provided
2. "backward": starts/ends the slice with the indices preceding the start/end time provided
3. "forward": starts/ends the slice with the indices following the start/end time provided

```python
import pynapple as nap
import numpy as np

ts = nap.Ts(t = np.arange(1, 5))
start, end = 2.2, 3.6

# slice over a range
print(ts.get_slice(start, end, mode="closest"))  # returns `slice(1, 3, None)`
print(ts.get_slice(start, end, mode="backward"))  # returns `slice(1, 2, None)`
print(ts.get_slice(start, end, mode="forward"))  # returns `slice(2, 3, None)`

# slice a single value
print(ts.get_slice(start, None, mode="closest"))  # returns `slice(1, 2, None)`
print(ts.get_slice(start, None, mode="backward")) # returns `slice(1, 2, None)`
print(ts.get_slice(start, None, mode="forward")). # returns `slice(2, 3, None)`
```

The "closest" mode is most similar to the behavior of pynapple "get" which finds the closest time point. 

The "backward" mode is useful, for example, if one wants to pick a frame from a video. Assume that the frame is displayed at time $[t_0, t_0 + dt] $, with 

`tsdtensor[tsdtensor.get_slice(t_0 + 3 * dt / 2, mode="backward")]` 

one will to pick the frame corresponding to $t_0$, while this would not be the case in the default mode.

@kushalkolar, @clewis7 this is supposed to be helpful for pynaviz, let me know if the slicing behavior is what you would expect/need!

Cheers,
Edo


pynapple/core/base_class.py
```diff
@@ -466,3 +466,162 @@ def _from_npz_reader(cls, file):
...
+        -------
+        slice : slice
+            A slice determining the start and end indices, with unit step.
+            - If mode = "closest":
+                Starts/ends the slice with indices closest to the start/end time provided.
+            - If mode = "backward":
+                Starts/ends the slice with the indices preceding the start/end time provided.
+            - If mode = "forward":
+                Starts/ends the slice with the indices following the start/end time provided.
+
```
gviejo, line 617: I would add an example section with the text of this PR conversation.

gviejo (changes requested):

pynapple/core/base_class.py
```diff
@@ -466,3 +466,179 @@ def _from_npz_reader(cls, file):
...
+        ValueError
+            - If start or end is not a number.
+            - If start is greater than end.
+
+        Examples
+        --------
+        >>> import pynapple as nap
+        >>> import numpy as np
+
+        >>> ts = nap.Ts(t = np.arange(1, 5))
```
gviejo, line 629:
```suggestion
        >>> ts = nap.Ts(t = [0,1,2,3])
```

pynapple/core/base_class.py
```diff
@@ -466,3 +466,179 @@ def _from_npz_reader(cls, file):
...
+        >>> start, end = 2.2, 3.6
```
gviejo, line 630:
```suggestion
        >>> start, end = 1.2, 2.6
```

gviejo (changes requested):

gviejo:
Thank you Edoardo :beer:

PR #323, opened 2024-07-30
Iset splitting

gviejo:
DOes this :

```
ep = nap.IntervalSet(0, 100)
ep.split(interval_size=1)
```
return a new IntervalSet of length 100 with intervals of size 1 second

### DIFF
--- a/tests/test_interval_set.py
+++ b/tests/test_interval_set.py
@@ -548,9 +548,8 @@ def test_split_errors():
     end = [5, 15, 20, 40]
     ep = nap.IntervalSet(start=start, end=end)
 
-    with pytest.raises(IOError) as e:
+    with pytest.raises(IOError,  match="Argument interval_size should of type float or int"):
         ep.split('a')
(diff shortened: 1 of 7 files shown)


tests/test_interval_set.py
```diff
@@ -511,8 +515,51 @@ def test_save_npz():
...
+    assert len(ep.split(1)) == 0
+
+def test_split_errors():
+    start = [0, 10, 16, 25]
+    end = [5, 15, 20, 40]
+    ep = nap.IntervalSet(start=start, end=end)
 
+    with pytest.raises(IOError) as e:
+        ep.split('a')
+    assert str(e.value) == "Argument interval_size should of type float or int"
```
BalzaniEdoardo, line 553:
```suggestion
    with pytest.raises(IOError,  match="Argument interval_size should of type float or int"):
        ep.split('a')
```

you can simplify the test if you want

BalzaniEdoardo (approved):
looks good, I suggested a slightly faster loop for splitting

PR #329, opened 2024-08-15
Filtering

BalzaniEdoardo:
I am starting off this as a draft PR since @gviejo and @BalzaniEdoardo are working together on it. 

The PR implements a wrapper scipy Butterworth filter, applicable to any `BaseTsd` object, correctly handling epoch boundaries.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,11 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main, dev ]
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
 
 jobs:
   lint:
--- a/pynapple/process/filtering.py
+++ b/pynapple/process/filtering.py
@@ -1,6 +1,7 @@
 """Filtering module."""
 
 import inspect
+from collections.abc import Iterable
 from functools import wraps
 from numbers import Number
 
@@ -18,11 +19,21 @@ def _validate_filtering_inputs(func):
         sig = inspect.signature(func)
         kwargs = sig.bind_partial(*args, **kwargs).arguments
 
-        if not isinstance(kwargs["cutoff"], Number):
-            if len(kwargs["cutoff"]) != 2 or not all(
-                isinstance(fq, Number) for fq in kwargs["cutoff"]
+        cutoff = kwargs["cutoff"]
+        filter_type = kwargs["filter_type"]
+        if filter_type in ["lowpass", "highpass"] and not isinstance(cutoff, Number):
+            raise ValueError(
+                f"{filter_type} filter require a single number. {cutoff} provided instead."
+            )
+        if filter_type in ["bandpass", "bandstop"]:
+            if (
+                not isinstance(cutoff, Iterable)
+                or len(cutoff) != 2
+                or not all(isinstance(fq, Number) for fq in cutoff)
             ):
-                raise ValueError
+                raise ValueError(
+                    f"{filter_type} filter require a tuple of two numbers. {cutoff} provided instead."
+                )
 
         if "fs" in kwargs:
             if kwargs["fs"] is not None and not isinstance(kwargs["fs"], Number):
@@ -218,12 +229,13 @@ def _compute_filter(
         raise ValueError("Unrecognized filter mode. Choose either 'butter' or 'sinc'")
 
 
-def compute_bandpass_filter(
+def apply_bandpass_filter(
     data, cutoff, fs=None, mode="butter", order=4, transition_bandwidth=0.02
 ):
     """
     Apply a band-pass filter to the provided signal.
     Mode can be :
+
         - 'butter' for Butterworth filter. In this case, `order` determines the order of the filter.
         - 'sinc' for Windowed-Sinc convolution. `transition_bandwidth` determines the transition bandwidth.
 
@@ -275,12 +287,13 @@ def compute_bandpass_filter(
     )
 
 
-def compute_bandstop_filter(
+def apply_bandstop_filter(
     data, cutoff, fs=None, mode="butter", order=4, transition_bandwidth=0.02
 ):
     """
     Apply a band-stop filter to the provided signal.
     Mode can be :
+
         - 'butter' for Butterworth filter. In this case, `order` determines the order of the filter.
         - 'sinc' for Windowed-Sinc convolution. `transition_bandwidth` determines the transition bandwidth.
 
@@ -332,12 +345,13 @@ def compute_bandstop_filter(
     )
 
 
-def compute_highpass_filter(
+def apply_highpass_filter(
     data, cutoff, fs=None, mode="butter", order=4, transition_bandwidth=0.02
 ):
     """
     Apply a high-pass filter to the provided signal.
     Mode can be :
+
         - 'butter' for Butterworth filter. In this case, `order` determines the order of the filter.
         - 'sinc' for Windowed-Sinc convolution. `transition_bandwidth` determines the transition bandwidth.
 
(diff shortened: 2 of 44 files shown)


gviejo (changes requested):

.github/workflows/main.yml
```diff
@@ -6,6 +6,11 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main, dev ]
+    types:
```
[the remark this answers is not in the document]
BalzaniEdoardo: Once merged this doesn't start actions in draft PRs, but only when the PR is ready

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
+    kwargs = dict(t=data.t, d=out, time_support=data.time_support)
+    if isinstance(data, nap.TsdFrame):
+        kwargs["columns"] = data.columns
+    return data.__class__(**kwargs)
+
+
+def _compute_spectral_inversion(kernel):
+    """
+    Compute the spectral inversion.
+    Parameters
```
BalzaniEdoardo, line 38:
```suggestion
    Compute the spectral inversion.
    
    Parameters
```

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
+    filter_type: str
+        Either 'lowpass', 'highpass', 'bandstop' or 'bandpass'.
+    transition_bandwidth: float
+        Percentage between 0 and 0.5
+    Returns
+    -------
+    np.ndarray
+    """
+    M = int(np.rint(20.0 / transition_bandwidth))
+    x = np.arange(-(M // 2), 1 + (M // 2))
```
BalzaniEdoardo, line 71:
```suggestion
    # This assures len(x) is odd, which is mandatory for spectral inversion to work.
    x = np.arange(-(M // 2), 1 + (M // 2))
```

I'll add a comment saying that you are making sure that you have an odd filter length because it is required for the filter to have the right property. I am worried that if we don't write it as a comment we will forget.

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
+    """
+    Compute the spectral inversion.
+    Parameters
+    ----------
+    kernel: ndarray
+
+    Returns
+    -------
+    ndarray
+    """
```
BalzaniEdoardo, line 45:
```suggestion
    """
    if len(kernel) % 2 == 0:
        raise ValueError("Spectral inversion requires an odd filter length.")
```

I know this is a private function, but I think a check here would make it the code more robust. If we edit our functions in the future, this is a good reminder

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
+    ----------
+    filter_type
+    """
+    kernel = _get_windowed_sinc_kernel(
+        freq / sampling_frequency, filter_type, transition_bandwidth
+    )
+    return data.convolve(kernel)
+
+
+def _validate_filtering_inputs(func):
```
BalzaniEdoardo, line 110: I probably would move this function at the top of the script.

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
+
+    Raises
+    ------
+    ValueError
+        If `data` is not a Tsd, TsdFrame, or TsdTensor.
+        If `cutoff` is not a tuple of two floats for "bandpass" and "bandstop" filters.
+        If `fs` is not float or None.
+        If `mode` is not "butter" or "sinc".
+        If `order` is not an int.
+        If "transition_bandwidth" is not a float.
```
BalzaniEdoardo, line 231:
```suggestion
        If "transition_bandwidth" is not a float.
        If `data` contains any NaNs.
```

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
```
BalzaniEdoardo, line 288: as before, add "if any nan are present.."

pynapple/process/filtering.py
```diff
@@ -0,0 +1,429 @@
...
+
+    Raises
+    ------
+    ValueError
+        If `data` is not a Tsd, TsdFrame, or TsdTensor.
+        If `cutoff` is not a number.
+        If `fs` is not float or None.
+        If `mode` is not "butter" or "sinc".
+        If `order` is not an int.
+        If "transition_bandwidth" is not a float.
```
BalzaniEdoardo, line 345: as before, add "if any nan are present.."

pynapple/process/filtering.py
```diff
@@ -0,0 +1,416 @@
...
```
BalzaniEdoardo, line 402: same as the other funcs

BalzaniEdoardo (commented):
This looks great. I think this module would be super useful

billbrod (commented):
I'm providing feedback on the new filtering tutorial, as well as the user interface.

User interface:
- it's a bit confusing to me that the functions are called `compute_*_filter`. I'm assuming it's because you have a lot of `compute_*` functions and want to be consistent, but you're not computing the filter, you're applying it to the signal. So calling it `apply_*_filter` or similar would make more sense to me.
- what's the use case for user setting `fs` themselves? it makes more sense to just grab that from the signal, since the user setting it themselves seems likely to lead to mistakes.
- As we've discussed for nemos, I don't love having a `mode` kwarg and then additional kwargs that are only used for one of the modes (`transition_bandwidth` and `order` here). I think an exception should be raised when they're set and shouldn't be, since that probably means a user has misunderstood something.
- are there allowable ranges for `order` and `transition_bandwidth`? Values that are so large and so small that they don't make sense? my guess is yes, and that at least a warning (if not an error) should be raised in those cases. (with a really large order, for example, you get overflow issues)
- `transition_bandwidth` has to be a float, but isn't parsing an int to a float reasonable here as well?
- The errors raised when `cutoff` is the wrong length is confusing. I think in the high-level function (e.g., `compute_highpass_filter`, not `_compute_filter`), the number of values of `cutoff` (1 or 2) should be explicitly checked and an informative `ValueError` raised. E.g., `highpass filters require a single cutoff value, but {len(cutoff)} values provided instead!`. (Among other things, the error message refers to `Wn`, and the user has no idea what that variable is)
- the docstrings should provide a basic description of what highpass/bandpass/lowpass/bandstop means. A one sentence description like, "preserves frequencies above the cutoff, discards those below".
- docstrings are not rendered correctly by mkdocs. there needs to be an extra newline between the first line and the one that starts `Mode can be :`

Tutorial:
- Could be nice to have some high-level description here: why is filtering used in neuroscience, with link to examples (including another of pynapple's tutorials!). And when would one use bandpass/lowpass/etc
- The question of how should one choose sinc vs. butter isn't addressed. they do give different outputs: is that important? or is the choice largely about efficiency?
- Same with how to choose the `cutoff`, `order`, and `transition_bandwidth` parameters. `cutoff` depends on the user's scientific question (but that should be said explicitly), and guidance on the others would be nice.
- when plotting the results of the bandpass filter, both filters reduce the amplitude of the signal  in the beginning. why is that? and why not show the full 2-second signal?
- also the x-axis on that plot is labelled `Time (Hz)`, it should be `Time (s)`, right?
- The frequency response plots are confusing -- took me a while to realize the top and bottom rows are the same plots, just with linear and log scales on the y-axis. it's also not clear what users should be taking from this plot.
    - To make it more useful: could show these filters applied to the signal (both the time and frequency domain representations) and talk through why you might want sharper / less sharp of a transition.
    - Is there a principled way to choose values for order / transition bandwidth, or is it empirical? you give some defaults, do those work in most cases or should people always experiment with them? 
- The reference to "kernel is too short" is confusing, since the user never sees the actual kernel. I would mention that in an aside (footnote or admonition) and point out what effect that has in the data (reduction of signal amplitude), and how the user should fix it (reduce transition bandwidth). 
    - This should be its own section, it's kind of "troubleshooting"
- The performance code block is really long and not actually interesting to the user. it would be nice to hide this. what we ended up [doing in nemos](https://github.com/flatironinstitute/nemos/pull/216) was moving the utility code to a hidden module. could do that or, since users probably won't run this themselves, could just move it to a utility script in the `docs` folder.
- that performance section should also interpret the output. why are there two plots here? what are we supposed to take from this?
- is it worth removing the noise from the example signal? or at least, do it with and without? real data always has noise, but it would make interpreting the bandstop filter easier if the noise wasn't present. the noise also means the true 10 Hz component and the outputs of the filters are different, which might confuse people. should at least explain this.

PR #330, opened 2024-08-19
Fixing nan bug when creating bins from min and max of feature

slcalgin:
Hello! 
I noticed that in functions such as `compute_1d_mutual_info()` (along with others), when the `minmax` argument is not provided, the bins are made using the minimum and maximum value of the provided feature using functions `np.min()` and `np.max()`. The issue with this, however, is that if there are any nans in the feature, then `np.min()` and `np.max()` will return nan; This results in bins of 'nans', and therefore mutual info values of all zeros. I've had a similar issue with tuning curves. To fix this, I changed the the `np.min()` and `np.max()` functions to `np.nanmin()` and `np.nanmax()` respectively. These functions ignore nans.

### DIFF
--- a/pynapple/process/tuning_curves.py
+++ b/pynapple/process/tuning_curves.py
@@ -119,7 +119,7 @@ def compute_1d_tuning_curves(group, feature, nb_bins, ep=None, minmax=None):
         assert isinstance(ep, nap.IntervalSet), "ep should be an IntervalSet"
 
     if minmax is None:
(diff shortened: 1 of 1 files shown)


gviejo:
I am refactoring the tuning curves module so I will merge Selen PR into this PR.

slcalgin:
Hi Guillaume, thanks for this--I didn't have a chance to come around to creating tests.

PR #364, opened 2024-11-06
nwb interface: allow full paths in __getitem__ (2)

magland:
This replaces https://github.com/pynapple-org/pynapple/pull/351


pynapple/io/interface_nwb.py
```diff
@@ -458,9 +458,27 @@ def __getitem__(self, key):
...
+            else:
+                original_key = None
             if self.__contains__(key):
                 if isinstance(self.data[key], dict) and "id" in self.data[key]:
-                    obj = self.nwb.objects[self.data[key]["id"]]
+                    obj_id = self.data[key]["id"]
+                    obj = self.nwb.objects[obj_id]
+                    if original_key is not None:
+                        if _path_for_nwb_object(obj) != original_key:
+                            raise KeyError(
```
gviejo, line 479: Is it possible to write a little tests for this? I am not sure I understand the case where this would happen.

BalzaniEdoardo (approved):
This looks good to me. I made a couple of edits:

1.  Improved the efficiency of the recursion unique path generation. The original recursion was always checking all the items in the dictionary, now at each recursion it only operates on the sub-items that have duplicate values.
2. Some stylistic changes on the iterate over nwb.
3. Parameterized the test for disambiguation of labels

PR #368, opened 2024-11-15
Update metadata docstrings

sjvenditto:
## Summary
This PR updates the docstrings for new metadata functions, as well as docstrings for classes that inherit metadata.
- all functions and properties in `metadata_class.py`, notably docstring examples added to `set_info` and `get_info` methods
- `interval_set.py > IntervalSet`, `time_series.py > TsdFrame`, and `ts_group.py > TsGroup`: doc strings that were previously split under class definition and init have been merged, and new docstring examples added for initializing each class with and without metadata
- docstrings were added to some properties throughout `IntervalSet`, `TsGroup`, `TsdFrame`, `_BaseTsd`, and `_Base`
- "Metadata" section on the "User Guide > Core methods" documentation page has been updated


doc/user_guide/04_core_methods.md
```diff
@@ -373,3 +267,256 @@ plt.xlabel("Time (s)")
...
+```
+
+### Metadata properties
+User-set metadata is mutable and can be overwritten.
+```{code-cell} ipython3
+tsgroup.set_info(region=["A","B","C"])
+print(tsgroup.get_info("region"))
+```
+
+If the metadata name does not overlap with an existing class column, it can be set and accessed via key indexing (i.e. using square brackets).
```
gviejo, line 494: Shouldn't this be in the section Accessing metadata?

gviejo: Also why mentioning class column only?

pynapple/core/metadata_class.py
gviejo, line 1: Ok something very annoying but for the docstrings of set_info, the examples appears as TsGroup for IntervalSet. Is there a way to have dynamical examples as a function of what is inherited?

pynapple/core/ts_group.py
gviejo, line 1: Ok here it's different from IntervalSet and TsdFrame. The init is basically empty of docstrings now. I think I prefer this way for the documentation.

pynapple/core/metadata_class.py
sjvenditto: This is a good idea, I'll look into it

pynapple/core/ts_group.py
sjvenditto: I think this comment in on an older version of `ts_group.py`, it should also have an empty init

pynapple/core/metadata_class.py
sjvenditto: added class-specific examples

pynapple/core/ts_group.py
sjvenditto: removed init strings from intervalset and tsdframe

doc/user_guide/04_core_methods.md
sjvenditto: I changed this to explicitly mention the exceptions for IntervalSet's "start" and "end"

gviejo (approved):

PR #388, opened 2025-01-14
Implementing build_tensor and warp_tensor in warping.py

gviejo:
Here is the proposed behavior : 

Simple trial-based tensor : 

```python
import pynapple as nap
import numpy as np
group = nap.TsGroup({0:nap.Ts(t=np.arange(0, 100))})
ep = nap.IntervalSet(start=np.arange(20, 100, 20), end=np.arange(20, 100, 20) + np.arange(2, 10, 2))
tensor = nap.build_tensor(group, ep, binsize=1)
print(tensor)
    array([[[ 1.,  1., nan, nan, nan, nan, nan, nan],
            [ 1.,  1.,  1.,  1., nan, nan, nan, nan],
            [ 1.,  1.,  1.,  1.,  1.,  1., nan, nan],
            [ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.]]])
```

Time-warped trial based tensor :

```python
import pynapple as nap
import numpy as np
group = nap.TsGroup({0:nap.Ts(t=np.arange(0, 100))})
ep = nap.IntervalSet(start=np.arange(20, 100, 20), end=np.arange(20, 100, 20) + np.arange(2, 10, 2))
tensor = nap.warp_tensor(group, ep, num_bin=10)
print(tensor)
    array([[[1., 0., 0., 0., 0., 1., 0., 0., 0., 0.],
            [1., 0., 1., 0., 0., 1., 0., 1., 0., 0.],
            [1., 1., 0., 1., 0., 1., 1., 0., 1., 0.],
            [1., 1., 1., 1., 0., 1., 1., 1., 1., 0.]]])
```

Both functions works for any time series.


sjvenditto:
I have some thoughts about the naming conventions used here:

1. Having `warp_tensor` as a separate function from `build_tensor` feels a little misleading since `warp_tensor` also "builds" a tensor. 

My thought is to have a central function `build_tensor` with a `method` parameter that can be `"pad"` or `"warp"`, that takes `method_kwargs`, so something like:
```
def build_tensor(input, ep, method, **method_kwargs)
```
and that way both ways of building a tensor can be accessed in a centralized location. In this way, the current `build_tensor` would be renamed to `pad_tensor`, and keep the kwargs definitions in the docstring. `warp_tensor` could remain as-is

2. Similar to numpy functions like `np.reshape`, I think it would be useful to define an object method `to_tensor` that calls `build_tensor`

What do you think about this?

gviejo:
I would have `build_tensor` as a method for Tsds and TsGroup but `warp_tensor` as a function of warping.py. I can see a path where more warping methods are added inflating the core while `build_tensor` with simple padding should remain small.

pynapple/core/time_series.py
```diff
@@ -638,6 +638,79 @@ def interpolate(self, ts, ep=None, left=None, right=None):
...
+        if align not in ["start", "end"]:
+            raise RuntimeError("align should be 'start' or 'end'")
+
+        slices = [self.get_slice(s, e) for s, e in ep.values]
+        lengths = list(map(lambda sl: sl.stop - sl.start, slices))
+        n_t = max(lengths)
+        output = np.ones(shape=(len(ep), n_t, *self.shape[1:])) * padding_value
+        if align == "start":
+            for i, sl in enumerate(slices):
+                output[i, 0 : lengths[i]] = self[sl].values
```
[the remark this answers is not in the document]
BalzaniEdoardo: %timeit tsd[sl].values
15.8 μs ± 1.1 μs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
%timeit tsd.values[sl]
75.9 ns ± 0.442 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

pynapple/core/ts_group.py
```diff
@@ -849,6 +850,109 @@ def to_tsd(self, *args):
 
         return toreturn
 
+    def to_trial_tensor(
```
BalzaniEdoardo, line 853: Discussed with @sjvenditto , maybe `count_trials` or `trials_count` to be more explicit about the operation

pynapple/process/perievent.py
```diff
@@ -1,12 +1,49 @@
...
+            "timeseries": (nap.Tsd, nap.TsdFrame, nap.TsdTensor),
+            "tref": (nap.Ts, nap.Tsd, nap.TsdFrame, nap.TsdTensor),
+            "group": (nap.TsGroup,),
+            "ep": (nap.IntervalSet,),
+            "feature": (nap.Tsd, nap.TsdFrame, nap.TsdTensor),
+            "binsize": (Number,),
+            "windowsize": (tuple, Number),
+            "time_unit": (str,),
+        }
+        for param, param_type in parameters_type.items():
```
BalzaniEdoardo, line 33: I don't know why this is in this pr, but can you make this more generic?
Like loop over the kwargs instead of the parameters_type (which is function specific) and if there is a match with `parameters_type` then check the type. 

In this way, if we have a consistent naming for the parameters all checks can be funneled to the decorator.

Additionally, we could help enforcing parameter name consistency: We could add a flag to the decorator saying `raise_if_missing` that trigger a raise an error when true whenever the decorated function is supposed to have canonical parameter names but it doesn't.

BalzaniEdoardo (changes requested):
The API looks great! Added some minor stuff mostly

pynapple/process/perievent.py
```diff
@@ -35,102 +72,105 @@ def _align_tsd(tsd, tref, window, time_support):
...
+    timestamps : Ts, Tsd, TsdFrame, TsdTensor or TsGroup
+        The timestamps to align to tref.
+        If Ts/Tsd/TsdFrame/TsdTensor, returns a TsGroup.
         If TsGroup, returns a dictionary of TsGroup
-    tref : Ts or Tsd
-        The timestamps of the event to align to
-    minmax : tuple, int or float
+    tref : Ts, Tsd, TsdFrame or TsdTensor.
+        The time reference of the event to align to
+    windowsize : tuple of int/float or int or float
```
sjvenditto, line 105: If we want to be consistent with how we format the `bin_size` parameter, we might want to change this to `window_size`

pynapple/process/perievent.py
```diff
@@ -188,22 +228,23 @@ def compute_perievent_continuous(data, tref, minmax, ep=None, time_unit="s"):
         return nap.TsdTensor(t=time_idx, d=new_data_array, time_support=time_support)
 
 
+@_validate_perievent_inputs
 def compute_event_trigger_average(
     group,
     feature,
     binsize,
```
sjvenditto, line 235: do we want to change this `binsize` to `bin_size` to match `count` (and now `trial_count`)?

pynapple/process/perievent.py
```diff
@@ -188,22 +228,23 @@ def compute_perievent_continuous(data, tref, minmax, ep=None, time_unit="s"):
...
-    windowsize=None,
+    windowsize=0,
     ep=None,
     time_unit="s",
 ):
     """
-    Bin the event timestamps within binsize and compute the Event Trigger Average (ETA) within windowsize.
+    Bin the event timestamps within binsize and compute the Event Trigger Average (ETA) within `windowsize`.
```
sjvenditto, line 241:
```suggestion
    Bin the event timestamps within binsize and compute the Event-Triggered Average (ETA) within `windowsize`.
```

pynapple/process/perievent.py
```diff
@@ -188,22 +228,23 @@ def compute_perievent_continuous(data, tref, minmax, ep=None, time_unit="s"):
         return nap.TsdTensor(t=time_idx, d=new_data_array, time_support=time_support)
 
 
+@_validate_perievent_inputs
 def compute_event_trigger_average(
```
sjvenditto, line 232: should we rename this to `compute_event_triggered_average`?

sjvenditto (changes requested):
I've committed the changes for changing some of the tensor function names that we discussed (also updated the tests and docs), and also pushed some grammatical fixes in the docs. 

I've suggested some things for the perievent changes, mostly with renaming parameters to homogenize the way we refer to things, specifically with renaming `windowsize` to `window_size`, and `binsize` to `bin_size`, to match other functions.

Most crucially, the user guide page for perievents needs to be updated with the parameter name changes.

pynapple/process/perievent.py
gviejo: Ok that would be a big breaking change. I would do it in another PR.

pynapple/process/perievent.py
gviejo: Yes. Let's keep that for another PR.

pynapple/core/time_series.py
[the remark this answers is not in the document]
gviejo: Done

pynapple/core/time_series.py
```diff
@@ -2223,3 +2296,73 @@ def save(self, filename):
...
+        Returns
+        -------
+        numpy.ndarray
+
+        Raises
+        ------
+        RuntimeError
+            If `time_unit` not in ["s", "ms", "us"]
+        """
+        if not isinstance(ep, IntervalSet):
```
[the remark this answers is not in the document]
gviejo: Ok not possible anymore since it's a different name now.

pynapple/process/perievent.py
gviejo: Yes

gviejo:
I reverted the argument name change in perievent because there was no easy way to raise a deprecation warning. Homogeneizing the argument name will have to be done for a major release.

PR #396, opened 2025-01-16
Hotfix - tsd initialization

BalzaniEdoardo:
Some of the methods of Tsd were not overwriting the epochs when `ep` was provided (like bin_average. or interpolate).
Fix the issue with this PR to fix the behavior of these operation. Added tests.


BalzaniEdoardo:
Compare the funtion I just fixed to this:

https://github.com/pynapple-org/pynapple/blob/d62403503eb0a27cf9a6c535fa50e3102cfbbce3/pynapple/core/time_series.py

which show how they were before the introduction of `_initialize_tsd_output`

gviejo (approved):

PR #406, opened 2025-01-30
Add ipython tab autocompletions for TsdFrame columns

eschombu:
Using [this stackoverflow answer](https://stackoverflow.com/a/58601400), we can get the behavior shown in the screenshot.
Screenshot 2025-01-30 at 1 00 55 PM


gviejo:
I am curious what it does when you don't have strings for columns

gviejo:
Ok I am not sure it's a good idea if you don't have labeled columns. It does this for integers but it's a bit misleading because this is numpy behavior now 

image

PR #411, opened 2025-02-05
Fix value printing of IntervalSet when rows are collapsed

sjvenditto:
The wrong values were getting printed on the bottom rows of the `__repr__` when rows are collapsed from a small terminal size


gviejo (approved):
ughhhh

PR #415, opened 2025-02-06
Overhaul of metadata to move away from pandas

sjvenditto:
Pandas DataFrames, while versatile, add a lot of overhead to object initialization with metadata, even when metadata is an empty DataFrame. Since slicing an existing object can often return a new object, this overhead is compounded each time an object is sliced.

In this PR, I've replaced the datatype of the private `_metadata` to a custom dictionary, where it previously was a pandas DataFrame. This custom dictionary includes minimal methods used by `_metadata`'s DataFrame counterpart -- e.g. `.loc`, `.iloc`, `.columns`, `.index` -- but is proving to be more lightweight. Rudimentary benchmarking suggests that, with the dictionary metadata, slicing `IntervalSet` and `TsdFrame` objects is 4-8X faster for objects with metadata and 2-3X faster for objects without metadata (i.e. empty metadata), when compared to objects with DataFrame metadata. (This speed up is not seen for `TsGroup` objects, which has a much slower initialization that the other objects, where metadata initialization is not the primary source of overhead)

On the user side of things, metadata will behave exactly the same as it did previously, where `obj.metadata` still returns a DataFrame.

## Class description
The private `_metadata`, handled by `_MetadataMixin`, is now a custom dictionary, `_Metadata` that inherits defaults from `UserDict` (e.g. `update`, `keys`, `items`, `values` methods). All functionality at the level of `_MetadataMixin` remains the same.

**Properties:** The `_Metadata` class has additional properties that mimic `pandas.DataFrame` to more easily allow for the object replacement.
- `columns`: the list of dictionary keys, or metadata labels
- `shape`: (length of index, length of columns)
- `dtypes`: data type for each metadata column
The more hardcore properties that have been re-implemented are `loc` and `iloc`, both of which use secondary classes to handle the indexing.
- `loc`: indexing by index value and column names, with two exceptions:
	- Boolean indexing is accepted in the first axis ONLY
	- `slice(None)` is accepted in the first axis ONLY
  This makes our version of `loc` slightly more restrictive than a `pandas.DataFrame`, which is more generous with slices and allows for boolean indexing and slices in the second axis.
- `iloc`: indexing by index and column position. Any indexing that `numpy` allows is allowed here. However, it treats tuples similar to `pandas` and not like `numpy`. For example, for a 2D `numpy.ndarray` array, `array[[0,1],[0,1]]` would return a 1D array with 2 values. For a 2D `pandas.DataFrame`, `df.iloc[[0,1],[0,1]]` would return a 2D array with 4 values. 

**Methods**: The class also includes additional methods for the same purpose. All these methods should be considered private and only used internally.
- `drop`: drop a column or columns from the metadata. operates in place. this is the "unsafe" method for dropping that is used to drop "rate" metadata from `TsGroup` objects internally
- `as_dataframe`: converts the metadata dictionary to a dataframe. this is called now by the public `metadata` attribute so that it still returns a dataframe.
- `groupby`: returns a dictionary of indices corresponding to unique values of a specific metadata column or columns
- `copy`: returns a deep copy of the metadata
- `reset_index`: resets index to integer values starting at 0
- `join`: column-wise join of metadata dictionaries. column names must be distinct between the two objects for join. only used by the `IntervalSet` method `join`
- `merge`: row-wise merge of metadata dictionaries. column names must match for two objects to be merged. the rows of the "other" dictionary are appended to the rows of the "base" dictionary. does not sort the resulting indices, where duplicate indices are possible. only used by the `TsGroup` method `merge_group`

Furthermore, `_Metadata` has custom implementations of `__repr__` and `__eq__`

**NOTES:** 
- The mechanics (e.g. .loc and .iloc) of `_Metadata` assumes/requires that metadata values are numpy arrays. This is NOT enforced, however, within `_Metadata`. I made this decision because `_Metadata` should not be created outside of `_MetadataMixin` except for testing, and `_MetadataMixin` converts all inputs to numpy arrays before being added to `_Metadata`. Therefore, I don't enforce this within `_Metadata` to reduce computation
- For length-1 objects, `_MetadataMixin` ensures that all metadata values are an array with length 1. Similarly, `.loc` and `.iloc` assume values are arrays for slicing. However, values *returned* by `.loc` and `.iloc` can break this assumption. This allows for simpler code with less length checks. All this really means is that `.loc` and `.iloc` cannot be chained on the private `_metadata` if it returns something with a single index, which is not something we do internally, nor is it something a user should be doing. They get turned back into arrays when its fed into a pynapple object.
## Additional notable changes
- a new function, `drop_info` was added to `_MetadataMixin` to drop metadata
- `get_info` was updated to only accept strings or lists of strings
- initialization of `TsGroup` was updated such that whenever the units' keys are sorted, metadata rows are also sorted to match
- `__repr__` methods for `IntervalSet`, `TsdFrame`, and `TsGroup` were all simplified (but look the same). `_convert_iter_to_str` needed to be updated as well
	- additionally, the `__repr__` for `TsdFrame` truncates metadata rows with a max of 3 rows
- `__getitem__` of `IntervalSet` was simplified to remove some special cases with metadata
- `__getitem__` of `TsdFrame` was simplified to no longer allow the `tsdframe[tsdframe.meta == x]` syntax. now metadata can only slice columns, e.g. `tsdframe[:, tsdframe.meta == x]`


pynapple/core/ts_group.py
```diff
@@ -721,8 +704,8 @@ def count(self, bin_size=None, ep=None, time_units="s", dtype=None):
                 )[1]
 
             metadata = self._metadata.copy()
-            # drop rate from metadata
-            metadata = metadata.drop(columns=["rate"])
+            # drop rate
+            metadata.drop("rate")
```
gviejo, line 708:
```suggestion
            metadata = metadata.drop(columns = ["rate"])
```
It's weird the test didn't catch it?

pynapple/core/ts_group.py
```diff
@@ -736,7 +719,7 @@ def count(self, bin_size=None, ep=None, time_units="s", dtype=None):
                 t=time_index,
                 d=np.empty((len(time_index), 0)),
                 time_support=ep,
-                metadata=self._metadata.drop(columns=["rate"]),
+                metadata=self._metadata.copy().drop("rate"),
```
gviejo, line 722: Probably not working here either. I have pandas '2.2.3'

pynapple/core/metadata_class.py
```diff
@@ -416,7 +447,383 @@ def anon_func(x):
...
+            raise TypeError(f"Unknown metadata index type {type(index)}")
+
+        if isinstance(columns, str):
+            data = {columns: self.data[columns][idx]}
+        else:
+            data = {k: self.data[k][idx] for k in columns}
+
+        return _Metadata(index, data)
+
+    def _get_indexder(self, vals):
```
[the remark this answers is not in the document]
sjvenditto: yes

PR #418, opened 2025-02-13
Set version dynamically

billbrod:
With this change, `setuptools_scm` will set the installed package version dynamically, based on the git tag. This is the same way I handle it for plenoptic and nemos (https://github.com/flatironinstitute/nemos/pull/297 and https://github.com/flatironinstitute/nemos/pull/308).

EDIT: The scheme will use the tag and then increment based on the number of commits since the git tag. When people install from pypi or from a github release, the version will be the same as the tag. If they install from a github branch that has additional commits, the patch version will increment and `devN` will be appended to it, where `N` is the number of commits since the last tag. For example, since the most recent tag is `v0.8.4`, using the current setup on this `dynamic_version` branch will give `0.8.5.dev10`. Thus, the only people who will see the `dev` tag are developers. 

EDIT 2: now using `no-guess-dev` scheme, so versions will look like `0.8.4.post1.dev10`

I don't *think* there's anything in pynapple that compares version numbers, but if you do, the proper way to do that is to use [`packaging.version.Version`](https://packaging.pypa.io/en/stable/version.html). I use this in [the plenoptic-documentation site](https://github.com/plenoptic-org/plenoptic-documentation/blob/43b8d2aac4a9dfddae4177bed29880b7e11253da/scripts/create_collection_md.py#L41) (and elsewhere) to sort version numbers.


billbrod:
Last commit removes `setup.py` and `setup.cfg`:
- `setup.py` was actually over-writing `pyproject.toml`, so that file was being ignored (which was not intended). I am able to install locally after removing `setup.py`, but let's double check whether that breaks anything else
- `setup.cfg` contained some tool config. The configuration for flake8 was moved to pyproject.toml (which required adding [flake8-pyproject](https://github.com/john-hen/Flake8-pyproject) to the devs dependencies, to enable this), while the one for `bdist_wheel: universal` was removed, because [it's deprecated](https://setuptools.pypa.io/en/latest/history.html#v75-1-0)
    - I just ran `python -m build --outdir dist/` locally, and the `tar.gz` and `pynapple-VERSION-py3-none-any.whl` file were both created, which are the [only files](https://pypi.org/project/pynapple/#files) you distribute.

gviejo (commented):
I find the .dev added to the local version very confusing. Is there an option within setuptools_scm to remove it or at least to not bump to a version that does not exist yet?

billbrod:
You can see the available choices [in the setuptools-scm docs](https://setuptools-scm.readthedocs.io/en/latest/extending/#version-number-construction). Here are some choices (all of this done with the head of this branch):
- `python-simplifier-semver` (choice currently in this PR): `0.8.5.dev11`
- `no-guess-dev`: `0.8.4.post1.dev11`
- `only-version`: `0.8.4`

So `only-version` matches your current behavior, and `no-guess-dev` doesn't increment the version number. I would not do the `only-version`, because it might be useful to make it obvious when someone is not on the release. For example, if someone installs from github and forgets about it (e.g., someone at one of our workshops), one of the other two schemes will make it clear that they have something other than a pypi release, which might make understanding bugs / unexpected behavior easier.

(as a note, [here's](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers) the details from the python packaging authority about the acceptable version strings)

pyproject.toml
```diff
@@ -35,6 +35,10 @@ requires-python = ">=3.8"
 [tool.setuptools.packages.find]
 include = ["pynapple", "pynapple.*"]
 
+[tool.setuptools_scm]
+version_scheme = 'python-simplified-semver'
```
gviejo, line 39:
```suggestion
version_scheme = 'no-guess-dev'
```

gviejo (approved):

PR #425, opened 2025-03-03
Publish workflow

billbrod:
This PR adds a new Github workflow file, `deploy.yml`, which will build and publish release to PyPI whenever you release them through Github. In order to make this work, you'll [need to set up trusted publishing](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/), which involves signing into PyPI and basically giving this github repo permission to upload for you (you'll need to specify the workflow's file name, `deploy.yml`).

In addition to building and deploying, this also checks:
- that we only built a single release
- that its version tag matches the most recent git tag
- that we can install from the wheel
- that the tests pass (right now, this is just `pytest tests/`, you could add other stuff if you like; for plenoptic, I also test that I can run the notebooks)

For [plenoptic](https://github.com/plenoptic-org/plenoptic/blob/main/.github/workflows/deploy.yml), I also add a manual `workflow_dispatch` trigger, which will cause the workflow to deploy to Test PyPI instead, just to be sure that everything works. If you'd like to do that as well, you'll need to get the pynapple name on TestPyPI (and also set up trusted publishing there). Nemos does not do that, and it's worked for nemos so far.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,73 @@
+name: Build and upload to PyPI for pure python
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    name: Build and test package
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+        # this is necessary for setuptools_scm to work properly with github
+        # actions, see https://github.com/pypa/setuptools_scm/issues/480 and
+        # https://stackoverflow.com/a/68959339
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
+    - name: Build package
+      run: |
+        pip install build
+        python -m build --outdir dist/ --sdist --wheel
+    - name: Check there's only one sdist and one whl file created
+      shell: bash
+        # because the following two tests will be weird otherwise. see
+        # https://askubuntu.com/a/454568 for why this is the right way to handle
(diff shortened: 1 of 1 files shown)


gviejo:
Does the tag have to be 1.1.1 or can it be v1.1.1? I usually label v1.1.1

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,73 @@
...
+        [[ $(find dist/ -type f -name "*whl" -printf x | wc -c) == 1 ]] || exit 1
+        [[ $(find dist/ -type f -name "*tar.gz" -printf x | wc -c) == 1 ]] || exit 1
+    - name: Check setuptools_scm version against git tag
+      shell: bash
+      run: |
+        # we use the error code of this comparison: =~ is bash's regex
+        # operator, so it checks whether the right side is contained in the
+        # left side. In particular, we succeed if the path of the source code
+        # ends in the most recent git tag, fail if it does not.
+        [[ "$(ls dist/*tar.gz)" =~ "-$(git describe --tags).tar.gz" ]]
```
gviejo, line 41:
```suggestion
        TAG=$(git describe --tags)
        [[ "$(ls dist/*tar.gz)" =~ "-${TAG:1}.tar.gz" ]]
```

gviejo: To remove the first letter

gviejo (changes requested):

billbrod:
Good catch on the first letter of the tag -- I don't include `v` in plenoptic's releases.

I checked locally, and I think everything should work as expected now.

PR #445, opened 2025-04-10
Preserve metadata when calling `count` on a `TsGroup`

sjvenditto:
This PR modifies the `TsGroup` `count` method to preserve metadata and pass it to the resulting `TsdFrame`. Additionally, it recomputes the "rate" metadata to be specific to the interval over which `count` was called.

I added checks for metadata to two tests in `test_ts_group.py`: `test_count` and `test_count_with_ep`. More generally, I refactored both of these tests to use parametrization to test more intervals and bin sizes, as well as checking the count results against similar numpy functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,4 @@ your
 
 # Ignore npz files from testing:
 tests/*.npz
+.vscode/settings.json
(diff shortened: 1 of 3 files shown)


BalzaniEdoardo (approved):
Looks good to me!

gviejo (approved):

Issue #389, opened 2025-01-15
Function to compute derivatives of time series


gviejo:
Loop over time support. Divided by bin interval

gviejo:
Done

PR #451, opened 2025-04-28
Add gradient method for calculating derivatives

jacobsimon:
Implements https://github.com/pynapple-org/pynapple/issues/389

## Summary

* Adds support for a new `derivative` method which wraps `np.gradient` with support for epochs and time index
* Adds unit tests with & without epochs

## Open questions
- [x] Do we want to call it something else like `grad` or `compute_derivatives` to distinguish from the original method?
- [x] Works with Tsd and TsdFrame along time axis - a little unclear to me if/how we should handle TsdTensor and whether we want to support `axis` or `edge_order` params 

## Example usage

Update: Added example to user docs:

Screenshot 2025-05-04 at 11 31 55 PM

Let me know what you think!

### DIFF
--- a/pynapple/core/time_series.py
+++ b/pynapple/core/time_series.py
@@ -638,6 +638,54 @@ class _BaseTsd(_Base, NDArrayOperatorsMixin, abc.ABC):
 
         return _initialize_tsd_output(self, new_d, time_index=new_t, time_support=ep)
 
+    def gradient_numpy(self, ts, ep=None, axis=None):
+        """Compute the gradient of the time series using numpy.gradient.
+
+        Parameters
+        ----------
+        ts : Ts, Tsd, TsdFrame or TsdTensor
+            The object holding the timestamps
+        ep : IntervalSet, optional
+            The epochs to use to interpolate. If None, the time support of Tsd is used.
+        axis : int, optional
+            Axis along which the gradient is computed. Default is None.
+
+        Returns
+        -------
+        Tsd, TsdFrame or TsdTensor
+            The gradient of the time series.
+        """
+        if not isinstance(ts, _Base):
+            raise IOError(
+                "First argument should be an instance of Ts, Tsd, TsdFrame or TsdTensor"
+            )
+
+        if ep is None:
+            ep = self.time_support
+        else:
+            if not isinstance(ep, IntervalSet):
+                raise IOError("ep should be an object of type IntervalSet")
+
+        new_t = ts.restrict(ep).index
+
+        new_shape = (
+            len(new_t) if self.values.ndim == 1 else (len(new_t),) + self.shape[1:]
+        )
+        new_d = np.full(new_shape, np.nan)
+
+        start = 0
+        print("EP", len(ep))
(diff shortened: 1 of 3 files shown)


pynapple/core/time_series.py
```diff
@@ -638,6 +638,51 @@ def interpolate(self, ts, ep=None, left=None, right=None):
 
         return _initialize_tsd_output(self, new_d, time_index=new_t, time_support=ep)
 
+    def gradient(self, ep=None):
+        """Compute the gradient of the time series using numpy.gradient.
```
gviejo, line 642: And then be super clear in the description that it computes the derivatives only in respect to time

tests/test_time_series.py
```diff
@@ -1157,6 +1157,49 @@ def test_interpolate_with_ep(self, tsd):
...
+        times = np.arange(0, 10, 0.001)
+
+        # Tsd
+        data = np.sin(times)
+        tsd = nap.Tsd(t=times, d=data)
+        expected_gradient = np.cos(times)  # Derivative of sin(x) is cos(x)
+        gradient = tsd.gradient()
+        np.testing.assert_allclose(gradient.values, expected_gradient, atol=1e-3)
+
+        # Tsdframe
```
gviejo, line 1174: This should be in `class TsdFrame` in test_time_series.py

gviejo (changes requested):

gviejo:
It's good thank you. There are just a few changes to do.
- A better name of the function would be `derivative` since we compute the gradient only in respect to time.
- Moving the test of `TsdFrame` to the class `TestTsdFrame`
- Adding the TsdTensor test
- Writing a docstring example would be nice. It can be very short. 
- It would be nice also to document it in the User guide more specifically [here](https://pynapple.org/user_guide/03_core_methods.html#time-series-method). You can write a little example and build the doc with sphinx-autobuild . _build/html` in the folder doc. The figure above can be shown there. Also I would use disjoint intervals to make it more clear that it jumps epochs.

pynapple/core/time_series.py
```diff
@@ -638,6 +638,51 @@ def interpolate(self, ts, ep=None, left=None, right=None):
 
         return _initialize_tsd_output(self, new_d, time_index=new_t, time_support=ep)
 
+    def gradient(self, ep=None):
```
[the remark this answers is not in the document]
jacobsimon: Updated in https://github.com/pynapple-org/pynapple/pull/451/commits/64946202de18f837247789d28f6c07ee519767c8

pynapple/core/time_series.py
jacobsimon: https://github.com/pynapple-org/pynapple/pull/451/commits/64946202de18f837247789d28f6c07ee519767c8

tests/test_time_series.py
jacobsimon: https://github.com/pynapple-org/pynapple/pull/451/commits/64946202de18f837247789d28f6c07ee519767c8

gviejo:
It's good thank you.

PR #462, opened 2025-05-30
Update docs and add contributor guide

jacobsimon:
## Summary

* Updates user guide and examples with links from class names and methods to the API reference 
* Adds a Makefile that supports `make lint`, `make test` and `make docs` (depends on `doc/Makefile`)
* Adds a simple `CONTRIBUTING` guide - loosely modeled after the `nemos` contributor guide


README.md
```diff
@@ -70,30 +70,18 @@ Getting Started
...
-or directly from the source code:
-
-``` {.sourceCode .shell}
-$ conda create --name pynapple pip python=3.8
-$ conda activate pynapple
-$ # clone the repository
-$ git clone https://github.com/pynapple-org/pynapple.git
-$ cd pynapple
-$ # Install in editable mode with `-e` or, equivalently, `--editable`
-$ pip install -e .
```
jacobsimon, line 91: Note: Moved this to the contributor guide because I don't think you'd usually install from source unless you're editing the library - but can move back if that's incorrect

gviejo: 👍

CONTRIBUTING.md
gviejo, line 1: I would reference it somewhere in the documentation. I would suggest adding it in the index.rst after citing?

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+git checkout -b your-branch-name
+
+# Commit and push changes
+git commit -m "Your commit message"
+git push origin your-branch-name
+```
+
+## Testing, linting, and building docs
+
+Helpful commands for running tests and generating documentation are defined in the root `Makefile`. These are generally kept in sync with the Github Actions defined in `.github/workflows`.
```
gviejo: To me this is what the tox file do. We could replace the tox file with a nox file eventually.

Makefile
gviejo, line 1: I think it's redundant with the tox file already there.

CONTRIBUTING.md
jacobsimon: Gotcha - I'll replace with instructions for using `tox` for now

CONTRIBUTING.md
jacobsimon: Ok I tried to add a link but had to remove it, because the doc linter was angry that the URL doesn't exist yet 😆  - can add in a later PR!

Issue #370, opened 2024-11-18
Implement ISI


gviejo:

wulfdewolf:
Some ideas/comments for implementing this:
- To my understanding, this comes down to `np.diff`, but applied to each interval separately. 
- We'd want a jitted version of that. However, there already exists [`jitdiff`](https://github.com/pynapple-org/pynapple/blob/22d6681ac5ccfc38911ab3b5a443b0a4dc88ab50/pynapple/core/_jitted_functions.py#L604), which is actually a `set_diff`, so I would propose to rename that to `jitset_diff`, matching the `IntervalSet` function [`set_diff`](https://github.com/pynapple-org/pynapple/blob/22d6681ac5ccfc38911ab3b5a443b0a4dc88ab50/pynapple/core/interval_set.py#L711) that uses it. Then `jitdiff` becomes available.
- wondering what the output would be? lists? dict? for a `Ts` it could be an array, but for a `TsGroup` the shapes won't be the same, so maybe a dict per unit?

wulfdewolf:
1. Makes sense! On it.
2. Right. A `nb_bins` argument then? Or a `bins` argument? 
2.1 For the log scale, a boolean argument `log_scale`? Do we want to allow different log bases?
2.2 For the name, `compute_isi_distribution` will indeed be most recognisable for neuroscience. Another possibility perhaps: `compute_time_diff_dist`, but also not great.

wulfdewolf:
I checked the improvement of a jitted version btw, not much indeed:
![Image](https://github.com/user-attachments/assets/a85ad586-0bc6-4003-9311-0f0712436156)
```
@jit(nopython=True, cache=True)
def jitdiff(time_array, starts, ends):
    n = len(time_array)
    m = len(starts)
    data = np.ones(n - 1, dtype=np.float64) * np.nan

    k = 0
    t = 0
    x = 0

    while ends[k] < time_array[t]:
        k += 1

    while k < m:
        # Outside
        while t < n:
            if time_array[t] >= starts[k]:
                break
            t += 1

        # Inside
        while t < n - 1:
            if time_array[t + 1] > ends[k]:
                k += 1
                break
            else:
                data[x] = time_array[t + 1] - time_array[t]
                x += 1
            t += 1

        if k == m:
            break
        if t >= n - 1:
            break
    return data[0:x]

```

PR #463, opened 2025-06-02
370 isi

wulfdewolf:
This addresses #370 by adding two things:
1. a `.time_diff()` method for time series objects
2. `nap.compute_isi_distribution`, which uses `.time_diff()` to compute the distribution of interspike intervals

In the docs, it changes `05_correlograms.md` to `05_discrete_events.md`, making logical space for `nap.compute_isi_distribution`.


pynapple/core/base_class.py
```diff
@@ -328,6 +338,56 @@ def count(self, bin_size=None, ep=None, time_units="s", dtype=None):
...
+
+        Parameters
+        ----------
+        align: str, optional
+            Determines the time index of the resulting time differences:
+             - "start" : the start of the interval between two timestamps.
+             - "center" [default]: the center of the interval between two timestamps.
+             - "end" : the end of the interval between two timestamps.
+        ep : IntervalSet, optional
+            The epochs to calculate time differences over. If None, the time support of the TsGroup is used.
```
sjvenditto:
```suggestion
            The epochs to calculate time differences over. If None, the time support of the object is used.
```

pynapple/process/correlograms.py
```diff
@@ -371,3 +371,66 @@ def compute_eventcorrelogram(
...
+    ep=None,
+):
+    """
+    Computes the interspike interval distribution.
+
+    Parameters
+    ----------
+    data : Ts, TsGroup, Tsd, TsdFrame or TsdTensor
+        The Ts, TsGroup, Tsd, TsdFrame or TsdTensor to compute the interspike interval distribution for.
+    nb_bins : int
```
sjvenditto: Should we also allow for custom bin edges?

tests/test_correlograms.py
```diff
@@ -519,3 +520,112 @@ def test_crosscorrelogram_reverse():
...
+            nap.Tsd(t=list(accumulate(range(11))), d=np.ones(11)),
+            10,
+            False,
+            pd.DataFrame(
+                index=np.arange(1.45, 10.45, 0.9), data=np.ones(10), columns=[0]
+            ),
+        ),
+    ],
+)
+def test_compute_isi_distribution(data, nb_bins, log_scale, expected):
```
sjvenditto: Also need tests for `TsdFrame` and `TsdTensor`

tests/test_correlograms.py
```diff
@@ -519,3 +520,112 @@ def test_crosscorrelogram_reverse():
...
+        ),
+        (
+            (get_group(), 2, True, []),
+            pytest.raises(
+                TypeError, match="ep should be an object of type IntervalSet"
+            ),
+        ),
+    ],
+)
+def test_compute_isi_distribution_type_errors(args, expectation):
```
sjvenditto, line 588: In addition to testing an empty list, I would also test variables that _could_ work in principle but should be rejected, e.g. `nb_bins = 2.0` and `ep = [0, 100]`

tests/test_correlograms.py
```diff
@@ -519,3 +520,112 @@ def test_crosscorrelogram_reverse():
...
+                data=np.ones(10),
+                columns=[0],
+            ),
+        ),
+        # TsGroup
+        (
+            nap.TsGroup(
+                {
+                    0: nap.Ts(list(accumulate(range(11)))),
+                    1: nap.Ts(list(accumulate(range(11)))),
```
sjvenditto: I would make sure the second unit has different spike times than the first. This test wouldn't catch the case where the result of the first unit is accidentally copied to every unit

tests/test_correlograms.py
```diff
@@ -519,3 +520,112 @@ def test_crosscorrelogram_reverse():
...
+        ),
+    ],
+)
+def test_compute_isi_distribution_type_errors(args, expectation):
+    with expectation:
+        nap.compute_isi_distribution(*args)
+
+
+@pytest.mark.parametrize(
+    "data, nb_bins, log_scale, expected",
```
sjvenditto: If you use separate `@pytest.mark.parametrize`'s for each parameter, you can test multiple permutations of parameter sets. At least, I would split out the `data` argument to reduce redundancy.

You can design each object to make the expected output the same in each test, and use if statements to handle exceptions (e.g. for `TsGroup`, for log scaled bins, or for 20 bins, etc). Then this can be defined inside the test to also reduce some redundancy

tests/test_time_series.py
```diff
@@ -2053,6 +2112,52 @@ def test_trial_count(self, ts):
...
+                "start",
+                nap.IntervalSet(start=[10, 50, 98], end=[20, 60, 100]),
+                nap.Tsd(
+                    d=np.ones(21),
+                    t=np.concatenate([np.arange(10, 20), np.arange(50, 60), [98]]),
+                ),
+            ),
+        ],
+    )
+    def test_time_diff(self, ts, align, ep, expected):
```
sjvenditto: This test is missing for `TsdFrame` and `TsdTensor`

I would move this test to `TestTimeSeriesGeneral` to easily catch all the object types. Since the timestamps should be the same in each of the objects, all the expected outputs will be the same for each input variation. This will reduce redundancy (right now it looks like the parametrization is exactly the same for `Tsd` and `Ts` objects)

doc/user_guide/03_core_methods.md
```diff
@@ -458,6 +458,39 @@ plt.xlabel("Trial time")
 plt.show()
 ```
 
+### `time_diff`
+
+`time_diff` computes the time differences between subsequent timepoints.
```
sjvenditto:
```suggestion
`Ts`, `Tsd`, `TsdFrame`, `TsdTensor`, and `TsGroup` all have the method `time_diff`, which computes the time differences between subsequent timepoints. For example, if a `Ts` object contained a set of spike times, `time_diff` would compute the inter-spike interval (ISI).

This method returns a new `Tsd` object, with values corresponding to each time difference, and time indices corresponding to the reference time point associated with each time difference. The reference time point can be adjusted by the optional `align` parameter, which can be set to  `"start"`, `"center"`, or `"end"` (the default being `"center"`).
```

doc/user_guide/03_core_methods.md
```diff
@@ -458,6 +458,39 @@ plt.xlabel("Trial time")
...
+```{code-cell} ipython3
+:tags: [hide-input]
+ts = nap.Ts(t=[1,5,6,12,16,18,19])
+```
+
+```{code-cell} ipython3
+time_diffs = ts.time_diff(align="center")
+print(time_diffs)
+```
+
```
sjvenditto:
```suggestion
Setting `align="center"` sets the reference time point to the midpoint between the timestamps used to calculate the time difference. Setting `align="start"` or `align="end"` sets the reference time point to the earlier or later timestamp, respectively.
```

sjvenditto: I would change the order of the plots to show "center" first since it is the default

doc/user_guide/03_core_methods.md
```diff
@@ -458,6 +458,39 @@ plt.xlabel("Trial time")
 plt.show()
 ```
 
+### `time_diff`
+
+`time_diff` computes the time differences between subsequent timepoints.
+The `align` parameter allows for specifying which timepoint should be used as index of the resulting `Tsd` (start, center or end).
```
sjvenditto:
```suggestion
```

doc/user_guide/05_discrete_events.md
```diff
@@ -71,3 +73,33 @@ eventcorrs = nap.compute_eventcorrelogram(
     )
 print(eventcorrs)
 ```
+
+## Interspike interval distribution (ISI)
```
sjvenditto:
```suggestion
## Interspike interval (ISI) distribution
```

doc/user_guide/05_discrete_events.md
```diff
@@ -71,3 +73,33 @@ eventcorrs = nap.compute_eventcorrelogram(
...
+
+The interspike interval distribution shows how the time differences between subsequent spikes (events) are distributed.
```
sjvenditto: I would add a sentence describing the required input parameters and a sentence describing the output of the function.

doc/user_guide/05_discrete_events.md
```diff
@@ -71,3 +73,33 @@ eventcorrs = nap.compute_eventcorrelogram(
...
+The `log_scale` parameter allows for using log space bins.
```
sjvenditto:
```suggestion
Setting the optional parameter `log_scale=True` allows for using log space bins.
```

I would put this sentence after the first example, and include a second example and plot showing the output using log-spaced bins

pynapple/core/ts_group.py
```diff
@@ -1022,6 +1022,56 @@ def trial_count(
...
+        5            6
+        8.5          1
+        dtype: float64, shape: (2,), 2: Time (s)
+        ----------  --
+        3.5          3
+        6            2
+        8            2
+        dtype: float64, shape: (3,)}
+        """
+        time_diffs = {}
```
sjvenditto:
```suggestion
        time_diffs = {k: v.time_diff(align=align, ep=ep) for k,v in self.data.items()}
```
I believe once `self.data` gets many keys, dictionary comprehension will beat out a for loop

pynapple/process/correlograms.py
```diff
@@ -371,3 +371,72 @@ def compute_eventcorrelogram(
...
+    if not isinstance(time_diffs, dict):
+        time_diffs = {0: time_diffs}
+
+    if log_scale:
+        time_diffs = {k: np.log(v) for k, v in time_diffs.items()}
+
+    if np.ndim(bins) == 0:
+        if bins < 1:
+            raise ValueError("`bins` must be positive, when an integer")
+        min_isi = min([isi for time_diff in time_diffs.values() for isi in time_diff])
```
sjvenditto:
```suggestion
        all_time_diffs = np.hstack([time_diff.d for time_diff in time_diffs.values()])
        min_isi, max_isi = min(all_time_diff), max(all_time_diffs)
```
this would reduce redundant for loops

pynapple/process/correlograms.py
```diff
@@ -371,3 +371,72 @@ def compute_eventcorrelogram(
...
+        bin_edges = np.asarray(bins)
+        if np.any(bin_edges[:-1] > bin_edges[1:]):
+            raise ValueError("`bins` must increase monotonically, when an array")
+    else:
+        raise ValueError("`bins` must be 1d, when an array")
+
+    bin_centers = (bin_edges[:-1] + bin_edges[1:]) / 2
+
+    counts = {}
+    for i in time_diffs:
```
sjvenditto: I would similarly replace with with dictionary comprehension

pynapple/core/ts_group.py
```diff
@@ -1022,6 +1022,56 @@ def trial_count(
...
+            Determines the time index of the resulting time differences:
+             - "start" : the start of the interval between two timestamps.
+             - "center" [default]: the center of the interval between two timestamps.
+             - "end" : the end of the interval between two timestamps.
+        ep : IntervalSet, optional
+            The epochs to calculate time differences over. If None, the time support of the TsGroup is used.
+
+        Returns
+        -------
+        dict
```
sjvenditto, line 1042: if you create a variable before returning, then this can be `time_diffs : dict`, but if you don't have a specific return variable than just `dict` is correct

tests/test_correlograms.py
```diff
@@ -519,3 +520,156 @@ def test_crosscorrelogram_reverse():
...
+)
+@pytest.mark.parametrize(
+    "bins",
+    [
+        1,
+        10,
+        np.linspace(0, 10, 10),
+        np.linspace(0, 2000, 100),
+        np.linspace(0, 2000, 1000),
+        np.geomspace(1, 100, 100),
```
sjvenditto, line 643: I would add a case where the bins sequence is a list as well

tests/test_correlograms.py
```diff
@@ -519,3 +520,156 @@ def test_crosscorrelogram_reverse():
...
+    ],
+)
+@pytest.mark.parametrize(
+    "ep",
+    [
+        None,
+        nap.IntervalSet([0, 10]),
+        nap.IntervalSet([0, 100]),
+        nap.IntervalSet([0, 2000]),
+        nap.IntervalSet([0, 4000]),
```
sjvenditto, line 653: I would add a case with multiple intervals in `ep`

pynapple/core/ts_group.py
```diff
@@ -1022,6 +1022,56 @@ def trial_count(
...
+        Computes the differences between subsequent timestamps.
+
+        Parameters
+        ----------
+        align: str, optional
+            Determines the time index of the resulting time differences:
+             - "start" : the start of the interval between two timestamps.
+             - "center" [default]: the center of the interval between two timestamps.
+             - "end" : the end of the interval between two timestamps.
+        ep : IntervalSet, optional
```
sjvenditto: This is annoying but we should replace all `ep` with `epochs` for the future standard. (This is especially annoying because you'll have to double check the docstring examples and error strings to make sure it all matches)

pynapple/process/correlograms.py
```diff
@@ -371,3 +371,72 @@ def compute_eventcorrelogram(
...
+    Parameters
+    ----------
+    data : Ts, TsGroup, Tsd, TsdFrame or TsdTensor
+        The Ts, TsGroup, Tsd, TsdFrame or TsdTensor to compute the interspike interval distribution for.
+    bins : int or sequence of scalars
+        If bins is an int, it defines the number of equal-width bins in the given range (10, by default).
+        If bins is a sequence, it defines a monotonically increasing array of bin edges, including the rightmost edge, allowing for non-uniform bin widths.
+    log_scale: bool, optional
+        If True, the computed ISI's are log-transformed. Default is False.
+    ep : IntervalSet, optional
```
sjvenditto: This also needs to change to `epochs`

tests/test_correlograms.py
```diff
@@ -519,3 +520,156 @@ def test_crosscorrelogram_reverse():
...
+)
+def test_compute_isi_distribution(data, bins, ep):
+    actual = nap.compute_isi_distribution(data, bins=bins, ep=ep)
+    assert isinstance(actual, pd.DataFrame)
+
+    time_diff = data.time_diff(ep=ep)
+    if not isinstance(time_diff, dict):
+        time_diff = {0: time_diff}
+    if isinstance(data, nap.TsGroup) and isinstance(bins, int):
+        min_isi = min([isi for isis in time_diff.values() for isi in isis])
```
sjvenditto, line 665: Another way to get the bins (if you don't want to copy the code from `compute_isi_distribution` itself) would be to stack all the values and use `np.histogram` to bin everything together first and get the expected edges. Not efficient in practice, but good for the test case to make numpy do the heavy lifting.

But also, if you update the code in `compute_isi_distribution` to what I suggested above, then this will already compute it in a different way from the function, which will also be a good sanity check

tests/test_time_series.py
```diff
@@ -922,6 +920,74 @@ def test_to_tensor(self, tsd):
...
+            ("center", None, nap.Tsd(d=np.ones(99), t=np.arange(0.5, 99.5))),
+            ("start", None, nap.Tsd(d=np.ones(99), t=np.arange(0, 99))),
+            # empty time support
+            (
+                "start",
+                nap.IntervalSet(start=[], end=[]),
+                nap.Tsd(d=[], t=[]),
+            ),
+            # empty epoch
+            (
```
sjvenditto, line 966: I can't think of a case where an IntervalSet with a single epoch would fail and an IntervalSet with multiple epochs would pass, but JUST in case, I would add a case with an IntervalSet with a single epoch (that's smaller than the time support). Otherwise, I really like the other cases that catch empty intervals and such

tests/test_ts_group.py
```diff
@@ -685,8 +683,143 @@ def test_trial_count_runtime_errors(
...
+                    2: nap.Tsd(d=np.full(1499, 0.2), t=np.arange(0.1, 299.8, 0.2)),
+                },
+            ),
+            # empty time support
+            (
+                "start",
+                nap.IntervalSet(start=[], end=[]),
+                {i: nap.Tsd(d=[], t=[]) for i in range(3)},
+            ),
+            # empty epoch
```
sjvenditto: also add the IntervalSet with one interval here too

doc/user_guide/01_introduction_to_pynapple.md
```diff
@@ -431,9 +431,9 @@ The `process` module of pynapple contains submodules that group methods that can
 Some functions have been doubled given the nature of the data. For instance, computing a 1d tuning curves from spiking activity requires the `nap.compute_1d_tuning_curves`. The same function for calcium imaging data which is a continuous time series is available with `nap.compute_1d_tuning_curves_continuous`. 
 :::
 
-**[Discrete correlograms](05_correlograms)**
+**[Discrete correlograms](05_correlograms_isi)**
 
-This module computes correlograms of discrete events, for example the cross-correlograms of a population of neurons.
+This module analyses discrete events, for example by computing the cross-correlograms of a population of neurons.
```
sjvenditto:
```suggestion
This module analyses discrete events, specifically correlograms (for example by computing the cross-correlograms of a population of neurons) and interspike interval (ISI) distributions.
```

doc/user_guide/01_introduction_to_pynapple.md
```diff
@@ -431,9 +431,9 @@ The `process` module of pynapple contains submodules that group methods that can
 Some functions have been doubled given the nature of the data. For instance, computing a 1d tuning curves from spiking activity requires the `nap.compute_1d_tuning_curves`. The same function for calcium imaging data which is a continuous time series is available with `nap.compute_1d_tuning_curves_continuous`. 
 :::
 
-**[Discrete correlograms](05_correlograms)**
+**[Discrete correlograms](05_correlograms_isi)**
```
sjvenditto:
```suggestion
**[Discrete correlograms & ISI](05_correlograms_isi)**
```

pynapple/core/ts_group.py
sjvenditto: we can also double check with Guillaume if he wants this done now

doc/user_guide/03_core_methods.md
```diff
@@ -458,6 +458,41 @@ plt.xlabel("Trial time")
 plt.show()
 ```
 
+### `time_diff`
+
+`Ts`, `Tsd`, `TsdFrame`, `TsdTensor`, and `TsGroup` all have the method `time_diff`, which computes the time differences between subsequent timepoints. For example, if a `Ts` object contained a set of spike times, `time_diff` would compute the inter-spike interval (ISI).
```
sjvenditto:
```suggestion
`Ts`, `Tsd`, `TsdFrame`, `TsdTensor`, and `TsGroup` all have the method `time_diff`, which computes the time differences between subsequent timepoints. For example, if a `Ts` object contained a set of spike times, `time_diff` would compute the interspike interval (ISI).
```

This is my bad, my suggestion from before used the hyphen but in general we are not hyphenating it

doc/user_guide/03_core_methods.md
```diff
@@ -458,6 +458,41 @@ plt.xlabel("Trial time")
...
+
+This method returns a new `Tsd` object, with values corresponding to each time difference, and time indices corresponding to the reference time point associated with each time difference. The reference time point can be adjusted by the optional `align` parameter, which can be set to  `"start"`, `"center"`, or `"end"` (the default being `"center"`).
```
sjvenditto:
```suggestion
This method returns a new `Tsd` object, with values being each time difference and time indices being their reference time point. The reference time point can be adjusted by the optional `align` parameter, which can be set to  `"start"`, `"center"`, or `"end"` (the default being `"center"`).
```
This is also my bad, I tend to be a little wordy. This sounds more concise (or also adjust it as you see fit)

doc/user_guide/03_core_methods.md
```diff
@@ -458,6 +458,41 @@ plt.xlabel("Trial time")
...
+    ax.set_title(f'ts.time_diff(align="{align}")')
+    if align != "end":
+        ax.set_xticks([])
+    for center, time_diff in zip(time_diffs.times(), time_diffs.values):
+        ax.plot([center, center], [-.25, time_diff], linestyle="--", c="black", zorder=-1)
+ax.set_xlabel("Time (s)")
+axs[0].legend(bbox_to_anchor=(1.05, 0.5, 0.5, 0.5))
+plt.show()
+```
+
```
sjvenditto, line 490: We also forgot a little blurb about the optional `ep` (maybe `epochs`) parameter. This is sort of inconsistently documented right now, maybe we just have something quick like "The optional parameter `epochs` can be passed to restrict where `time_diff` is calculated"

doc/user_guide/05_correlograms_isi.md
```diff
@@ -0,0 +1,156 @@
...
+eventcorrs = nap.compute_eventcorrelogram(
+    group=ts_group, event = nap.Ts(t=[0, 10, 20]), binsize=0.1, windowsize=1
+    )
+print(eventcorrs)
+```
+
+## Interspike interval (ISI) distribution
+
+The interspike interval distribution shows how the time differences between subsequent spikes (events) are distributed.
+The input can be any object with timestamps. Passing an `ep` restricts the computation to the given epoch.
```
sjvenditto: This looks good, just flagging the `ep` in case we change it

sjvenditto (changes requested):
Great job with the change to the new `bins` parameter! I just noted a couple more test cases to add in and a few final code and docs suggestions. The most annoying thing is if we want to rename all `ep` to `epochs`. Otherwise this PR should be good to go after these final comments!

pynapple/process/correlograms.py
```diff
@@ -371,3 +371,69 @@ def compute_eventcorrelogram(
...
+        If True, the computed ISI's are log-transformed. Default is False.
+    epochs : IntervalSet, optional
+        The epochs on which interspike intervals are computed.
+        If None, the time support of the input is used.
+
+    Returns
+    -------
+    pandas.DataFrame
+        DataFrame to hold the distribution data.
+
```
sjvenditto, line 402: We should have a quick docstring example here (it can be the same as what's in the docs now) (we need to revisit this module eventually to add docstring examples to the other functions too)

sjvenditto (changes requested):
One last tiny thing I just noticed -- `compute_isi_distribution` is missing examples. Otherwise everything else is perfect

Issue #475, opened 2025-06-26
Generalising `decode_1d` and `decode_2d` to `decode`


wulfdewolf:
We want to have one `decode` function that works for n-dimensional tuning curves passed.
The current machinery generalises to n dimensions, we just need to add some checks on the shape of the passed tuning curves.

wulfdewolf:
Closed, see #480

PR #478, opened 2025-07-03
477 n-dimensional tuning curves

wulfdewolf:
This combines:
- `compute_1d_tuning_curves`
- `compute_1d_tuning_curves_continuous`
- `compute_2d_tuning_curves`
- `compute_2d_tuning_curves_continuous`

into a generalised n-dimensional `compute_tuning_curves`, which now returns an `xarray.DataArray` (new dependency).

The old functions are still available, but they call the new one under the hood and show a `DeprecationWarning`.
A couple of notes (these clarify why I had to make some changes to the old tuning curve function tests):

- for `compute_1d_tuning_curves`, we were using `features.rate` to scale the tuning curves to Hz, but if one passes `epochs` that are bigger than that of the input features, this results in an incorrect computation of the rate and thus incorrect tuning curves.
FIXES: we now compute the rate as `1/np.mean(features.time_diff(epochs))`. 
There further is an optional `fs` argument, for if you want to pass the exact sampling rate yourself.

- for `compute_1d_tuning_curves_continuous` and `compute_2d_tuning_curves_continuous`, we were using `np.digitize` to bin, but this does not include the rightmost edge , unlike `np.histogram`.
FIXES: we now use `np.histogram` for all tuning curves. 
Hence, the new function is more consistent, as all tuning curves will now include the rightmost edge.

I have further updated all the notebooks and docs to use the generalised function.
The new `xarray.DataArray` output allows for some more concise plotting in a lot of the notebooks.

At the moment, the decoding is still separated into 1D and 2D, generalising that will be another [PR](https://github.com/pynapple-org/pynapple/issues/475).
For now, I am transforming the `xarray.DataArrays` into the respective old formats, such that the decoding functions still work.

### DIFF
--- a/pynapple/process/tuning_curves.py
+++ b/pynapple/process/tuning_curves.py
@@ -257,7 +257,9 @@ def compute_tuning_curves(
             or not all(isinstance(n, str) for n in feature_names)
         ):
             raise TypeError("feature_names should be a list of strings.")
-        if len(feature_names) != features.shape[1]:
+        if len(feature_names) != (
+            1 if isinstance(features, nap.Tsd) else features.shape[-1]
+        ):
             raise ValueError("feature_names should match the number of features.")
 
     # check epochs


pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+        If not passed, the column names in `features` are used.
+        If those are not set, they are set to `feature0`, `feature1`, etc.
+        You can also pass a list of tuples [(name, unit), ...] to set both names
+        and units in the resulting xarray.DataArray.
+
+    Returns
+    -------
+    xarray.DataArray
+        An xarray.DataArray containing the tuning curves with labeled dimensions.
+    """
```
sjvenditto, line 238: I would add in some examples

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+        epochs = features.time_support
+    elif isinstance(epochs, nap.IntervalSet):
+        features = features.restrict(epochs)
+    else:
+        raise TypeError("epochs should be an IntervalSet.")
+    group = group.restrict(epochs)
+
+    # check fs
+    if fs is None:
+        fs = 1 / np.mean(features.time_diff(epochs=epochs).values)
```
sjvenditto, line 276: Is this more accurate than `features.rate`?

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+    feature_names : sequence, optional
+        A sequence of names (and optionally units) for the features.
+        If not passed, the column names in `features` are used.
+        If those are not set, they are set to `feature0`, `feature1`, etc.
+        You can also pass a list of tuples [(name, unit), ...] to set both names
+        and units in the resulting xarray.DataArray.
+
+    Returns
+    -------
+    xarray.DataArray
```
sjvenditto, line 132: I think we also want to document that `occupancy` and `bin_edges` are returned as attributes as well

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+        raise TypeError("features should be a Tsd or TsdFrame.")
+
+    # check feature names
+    if feature_names is None:
+        _feature_names = features.columns
+        _feature_units = [None] * len(_feature_names)
+    else:
+        if not isinstance(feature_names, list):
+            raise TypeError("feature_names should be a sequence of strings or tuples.")
+        if len(feature_names) != features.shape[1]:
```
sjvenditto: alternatively, you could convert `feature_names` to a `numpy.array`, and check the shape in the first dimension
- if `ndims == 1`, then its the feature names only, 
- if `ndims == 2`, then `feature_names[:,0]` is the names, `feature_names[:,1]` is the units

This can also let you check `dtype` to make sure the contents are strings for both cases

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+            raise ValueError("feature_names should match the number of features.")
+        _feature_names = []
+        _feature_units = []
+        for feature in feature_names:
+            if isinstance(feature, str):
+                _feature_names.append(feature)
+                _feature_units.append(None)
+            elif isinstance(feature, tuple) and len(feature) == 2:
```
sjvenditto: this currently isn't enforcing that the feature name and units are strings (see above for potential fix)

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+            tcs[i] = np.histogramdd(
+                values,
+                weights=group.values[:, i],
+                bins=bin_edges,
+            )[0]
+        tcs /= counts
+        tcs[np.isnan(tcs)] = 0.0
+        tcs[:, occupancy == 0.0] = np.nan
+
+    return xr.DataArray(
```
sjvenditto: I think we want the 1D case to still return a pandas dataframe

tests/test_tuning_curves.py
```diff
@@ -736,13 +496,13 @@ def test_compute_1d_tuning_curves_continuous(args, kwargs, expected):
             ),
             2,
             {},
-            {"x": np.array([[1, 0], [0, 0]]), "y": np.array([[2, 0], [0, 0]])},
+            {"x": np.ones((2, 2)), "y": np.ones((2, 2)) * 2},
```
sjvenditto, line 1118: why is the expected output changing in this test? (not that I'm saying it's wrong now, more just curious what changed)

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+    elif not isinstance(features, nap.TsdFrame):
+        raise TypeError("features should be a Tsd or TsdFrame.")
+
+    # check feature names
+    if feature_names is None:
+        _feature_names = features.columns
+        _feature_units = [None] * len(_feature_names)
+    else:
+        if not isinstance(feature_names, list):
+            raise TypeError("feature_names should be a sequence of strings or tuples.")
```
sjvenditto:
```suggestion
            raise TypeError("feature_names should be a list of strings or tuples.")
```

sjvenditto: to be more explicit

tests/test_tuning_curves_general.py
```diff
@@ -0,0 +1,587 @@
...
+            {"feature_names": [1]},
+            pytest.raises(
+                TypeError,
+                match="feature_names should be a sequence of strings or tuples of strings.",
+            ),
+        ),
+        (
+            get_group(1),
+            get_features(1),
+            {"feature_names": [(1,)]},
```
sjvenditto: the test case `feature_names=[(1,1)]` does not currently fail

sjvenditto: After addressing this case, I would also test for like `("feature0", 1)` and `(1, "x")` as well, as these should also fail

tests/test_tuning_curves_general.py
```diff
@@ -0,0 +1,587 @@
...
+        (
+            get_group(1),
+            get_features(1),
+            {"feature_names": [("feature0", "x")]},
+            does_not_raise(),
+        ),
+        (
+            get_group(1),
+            get_features(2),
+            {"feature_names": [("feature0", "x"), "feature1"]},
```
sjvenditto: My `numpy` suggestion above would not work for this case, so it's a matter of if we want this flexibility

pynapple/process/tuning_curves.py
```diff
@@ -81,6 +82,259 @@ def wrapper(*args, **kwargs):
...
+
+    Parameters
+    ----------
+    group : TsGroup, TsdFrame or dict of Ts, Tsd objects.
+        The group of Ts or Tsd for which the tuning curves will be computed
+    features : Tsd, TsdFrame
+        The features (i.e. one column per feature).
+    bins : sequence or int
+        The bin specification:
+
```
sjvenditto, line 105:
```suggestion
```

tests/test_tuning_curves_general.py
```diff
@@ -0,0 +1,587 @@
+"""Tests of tuning curves for `pynapple` package."""
```
sjvenditto: This is a small nitpick, but we can probably combine these tests into the `test_tuning_curves.py` file. You probably have to rename the new defs to `get_group_n` and `get_features_n` or something

doc/user_guide/06_tuning_curves.md
```diff
@@ -64,19 +60,15 @@ The value is the mean firing rate of the neuron during this set of intervals.
...
 ```{code-cell} ipython3
 mean_fr = nap.compute_discrete_tuning_curves(tsgroup, dict_ep)
-
 pprint(mean_fr)
 ```
 
-
 ## From timestamps activity
   
-### 1-dimension tuning curves
```
sjvenditto, line 74: I think we should have a single big section now for all n-dimensional tuning curves, with a blurb at the top going through the inputs and outputs (since now they're the same for all the examples below). Then, I would have the 4 examples use cases as subsections (1d, 2d from timestamps, 1d, 2d from continuous)

doc/examples/tutorial_HD_dataset.md
```diff
@@ -73,108 +70,100 @@ print(data)
...
-```{code-cell} ipython3 
+```{code-cell} ipython3
 print(spikes)
 ```
 
 Here, rate is the mean firing rate of the unit. Location indicates the brain region the unit was recorded from, and group refers to the shank number on which the cell was located.
 
-This dataset contains units recorded from the anterior thalamus. Head-direction (HD) cells are found in the anterodorsal nucleus of the thalamus (henceforth referred to as ADn). Units were also recorded from nearby thalamic nuclei in this animal. For the purposes of our tutorial, we are interested in the units recorded in ADn. We can restrict ourselves to analysis of these units rather easily, using Pynapple.
-
+This dataset contains units recorded from the anterior thalamus. Head-direction (HD) cells are found in the anterodorsal nucleus of the thalamus (henceforth referred to as ADn).
```
sjvenditto: Is your linter making this multi-line? On my rendered version, it's splitting all the sentences to new lines too, although I'm not sure why... I thought it was supposed to be interpreted as a single paragraph until a line break?
image

My best guess is that these lines have double trailing spaces which may be causing the line break? (Vs. the paragraph below "Decoding", which seems to be rendered normally)

doc/examples/tutorial_HD_dataset.md
```diff
@@ -183,36 +172,38 @@ plt.show()
...
 
 
-+++
-
 ***
 Decoding
 ------------------
 
-Now that we have HD tuning curves, we can go one step further. Using only the population activity of ADn units, we can decode the direction the animal is looking in. We will then compare this to the real head direction of the animal, and discover that population activity in the ADn indeed codes for HD.
+Now that we have HD tuning curves, we can go one step further. Using only the population activity of ADn units, we can decode the direction the animal is looking in.
```
sjvenditto, line 179: This one here is rendering like a normal paragraph.

image

My best guess is that this is because line 179 appears to have a single trailing space?

pynapple/process/tuning_curves.py
wulfdewolf: I think so! This is something I discussed with Guillaume, because `.rate` uses the time support to compute the rate.
If someone passes a time support that is bigger than the actual time support, you're artificially inflating things.

tests/test_tuning_curves.py
wulfdewolf: Yes so, I wrote the explanations for those above in the PR description (apologies it was out-of-date before).
In short: we now use `np.histogram` instead of `np.digitize`, which doesn't include the last bin edge.

pynapple/process/tuning_curves.py
```diff
@@ -81,197 +82,413 @@ def wrapper(*args, **kwargs):
...
+            tcs[i] = np.histogramdd(
+                values,
+                weights=group.values[:, i],
+                bins=bin_edges,
+            )[0]
+        tcs /= counts
+        tcs[np.isnan(tcs)] = 0.0
+        tcs[:, occupancy == 0.0] = np.nan
+
+    if return_pandas and features.shape[1] == 1:
```
sjvenditto: One thing we could also do is first make the xarray, and then if `return_pandas=True`, it returns `tcs.to_pandas().T`. That way it will also throw an error if someone specifies `return_pandas=True` with 2 or more features, in case they're like "why didn't this give me a pandas dataframe"

pynapple/process/tuning_curves.py
```diff
@@ -81,197 +82,413 @@ def wrapper(*args, **kwargs):
...
+            attrs={"occupancy": occupancy, "bin_edges": bin_edges},
+        )
 
 
 @_validate_tuning_inputs
 def compute_1d_tuning_curves(group, feature, nb_bins, ep=None, minmax=None):
-    """
-    Computes 1-dimensional tuning curves relative to a 1d feature.
+    warnings.warn(
+        "compute_1d_tuning_curves is deprecated and will be removed in v1.0; "
```
sjvenditto:
```suggestion
        "compute_1d_tuning_curves is deprecated and will be removed in a future version; "
```

it will probably be 1.0, but we might want to double check with Guillaume or be less specific

tests/test_tuning_curves.py
```diff
@@ -817,3 +827,351 @@ def test_compute_2d_tuning_curves_continuous(tsdframe, nb_bins, kwargs, expected
     for i in tc.keys():
         assert tc[i].shape == nb_bins
         np.testing.assert_almost_equal(tc[i], expected[i])
+
```
sjvenditto: I would put a comment block here akin to lines 14 and 433 just to demarcate that these are general tuning curve tests

pynapple/process/tuning_curves.py
```diff
@@ -81,197 +82,413 @@ def wrapper(*args, **kwargs):
...
+        )
+    elif not isinstance(features, nap.TsdFrame):
+        raise TypeError("features should be a Tsd or TsdFrame.")
 
-        tuning_curves.loc[k] = tuning_curves.loc[k] / dict_ep[k].tot_length("s")
+    # check feature names
+    if feature_names is None:
+        feature_names = features.columns
+    else:
+        if not isinstance(feature_names, list) or not all(
```
sjvenditto: we could technically be a little more inclusive with this, i.e. do a check like
```suggestion
        if not hasattr(feature_names, "len") or not all(
```
one annoying this is that strings pass the first check, and would pass the second check if the number of characters is the same as the number of features.

also, do we want to allow for a string to be passed if there is a single feature?

tests/test_tuning_curves.py
```diff
@@ -256,168 +197,236 @@ def test_compute_2d_mutual_info_errors(
...
+        (
+            get_group_n(1),
+            get_features_n(1),
+            {"feature_names": ["feature0", "feature1"]},
+            pytest.raises(
+                ValueError, match="feature_names should match the number of features."
+            ),
+        ),
+        (
+            get_group_n(1),
```
sjvenditto, line 201: whether or not we want to be inclusive for other array-like objects, we should test for passing feature names as something like an `np.array` or a `tuple`

sjvenditto (commented):
I've left a few final comments that are more inconsequential, so let me know your thoughts on if you want to put in these small changes -- otherwise, this one's just about done!

pynapple/process/tuning_curves.py
wulfdewolf: Hmm for something like column names not allowing just a string feels better to me.
For `bins` it makes sense to allow a tuple vs. a list of tuple(s), since it is applied to all dimensions?
It feels very much like setting columns in a `pd.DataFrame` to me, which I would never expect to do with just a string, even if there is only one column?

wulfdewolf: I made it this for now:
```
        if (
            not hasattr(feature_names, "__len__")
            or isinstance(feature_names, str)
            or not all(isinstance(n, str) for n in feature_names)
        ):
```

sjvenditto: That sounds good to me!

sjvenditto (approved):

main.py
```diff
@@ -16,16 +16,18 @@
...
 # COMPUTING TUNING CURVES
-tuning_curves = nap.compute_1d_tuning_curves(
-    spikes, head_direction, 120, minmax=(0, 2 * np.pi)
+tuning_curves = nap.compute_tuning_curves(
+    spikes,
+    head_direction,
+    120,
+    epochs=wake_ep,
+    range=(0, 2 * np.pi),
+    feature_names=[("head direction", "rad")],
```
gviejo: This should be `feature_names=["head direction"],`

pynapple/process/tuning_curves.py
gviejo, line 1: Do you want to update the header of the file to reflect the changes?

doc/examples/tutorial_calcium_imaging.md
gviejo, line 1: Do you want to append your name at the end?

gviejo (approved):

PR #480, opened 2025-07-10
475 decode generalisation

wulfdewolf:
This combines:
- `decode_1d`
- `decode_2d`

into a generalised n-dimensional `decode_bayes`, which takes as input an `xarray.DataArray` (the new output of `compute_tuning_curves`).

Since `occupancy` is now an attribute of the tuning curves, `decode_bayes` uses that directly, if a new argument `use_occupancy` is set to `True`, it is `False` by default.

(this builds on the [generalised tuning curves PR](https://github.com/pynapple-org/pynapple/pull/478)) 

The old functions (and tests) are still available but they throw a `DeprecationWarning`.


wulfdewolf:
@gviejo we need your opinion on arguments for `nap.decode_bayes`:
since the occupancy is now an attribute in the input `tuning_curves`, @sjvenditto & I thought we should add a `use_occupancy` argument that is `False` by default but can be set to `True` if the user wants it. 

Do you think the default should be `False`?
Is the argument name good?

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
+        The epochs on which decoding is computed
     bin_size : float
         Bin size. Default is second. Use the parameter time_units to change it.
     time_units : str, optional
         Time unit of the bin size ('s' [default], 'ms', 'us').
-    feature : Tsd, optional
-        The 1d feature used to compute the tuning curves. Used to correct for occupancy.
-        If feature is not passed, the occupancy is uniform.
+    use_occupancy : bool, optional
+        If True, uses the occupancy from the tuning curves to adjust the probability distribution.
```
sjvenditto:
```suggestion
        If True, uses the occupancy from the tuning curves as a prior over the feature probability distribution.
```

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
+        If False, uses a uniform distribution. Default is False.
```
sjvenditto:
```suggestion
        If False, uses a uniform distribution as a prior. Default is False.
```

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
-        t=count.index, d=p, time_support=ep, columns=tuning_curves.index.values
-    )
-
-    decoded = nap.Tsd(
-        t=count.index, d=tuning_curves.index.values[idxmax], time_support=ep
+    p = p.reshape(p.shape[0], *tuning_curves.shape[1:])
+    p = getattr(nap, f"Tsd{'Tensor' if p.ndim > 2 else 'Frame'}")(
+        t=count.index,
+        d=p,
+        time_support=epochs,
```
sjvenditto: For a `TsdFrame`, we also want the columns names to be set to the feature values

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
+    95.5        1.0  1.0
+    96.5        1.0  1.0
+    97.5        1.0  1.0
+    98.5        1.0  1.0
+    dtype: float64, shape: (98, 2)
+    """
 
-        count = group
+    # check tuning curves
+    if not isinstance(tuning_curves, xr.DataArray):
```
sjvenditto, line 194: We might want to consider allowing a pandas DataFrame in the 1D case, which will either only work for `use_occupancy=False`, or we would have an optional parameter `occupancy` to pass it as well

I don't feel strongly about this though if we want to enforce the xarray

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
-        raise RuntimeError("Unknown format for feature in decode_1d")
+        raise TypeError("Unknown format for group.")
+
+    if use_occupancy:
+        if "occupancy" not in tuning_curves.attrs:
+            raise ValueError(
+                "use_occupancy set to True but no occupancy found in tuning curves."
+            )
+        occupancy = tuning_curves.attrs["occupancy"]
+        if occupancy.shape != tuning_curves.shape[1:]:
```
sjvenditto: would this ever happen when the occupancy is computed with the tuning curves?

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
+    tuning_curves : xr.DataArray
+        Tuning curves as outputed by `compute_tuning_curves` (one for each unit).
+    group : TsGroup, TsdFrame or dict of Ts, Tsd
+        A group of neurons with the same keys as the tuning curves.
         You may also pass a TsdFrame with smoothed rates (recommended).
-    ep : IntervalSet
-        The epoch on which decoding is computed
+    epochs : IntervalSet
+        The epochs on which decoding is computed
     bin_size : float
```
sjvenditto, line 35: This might be beyond the scope of this PR since it is also dependent on `count`, but I've always wanted to be able to specify custom bins, or have non-uniform bin sizes across epochs/trials. Whether this should be implemented / how to implement it well would require some discussion

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
-        count = newgroup.count(bin_size, ep, time_units)
+    return decoded, p
 
-    elif type(group) is dict:
-        newgroup = nap.TsGroup(group, time_support=ep)
-        count = newgroup.count(bin_size, ep, time_units)
 
+def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
+    warnings.warn(
+        "decode_1d is deprecated and will be removed in v1.0; use decode instead.",
```
sjvenditto:
```suggestion
        "decode_1d is deprecated and will be removed in a future version; use decode instead.",
```

pynapple/process/decoding.py
```diff
@@ -19,194 +25,308 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
+        bin_size,
+        time_units=time_units,
+        use_occupancy=feature is not None,
+    )
 
-    indexes = list(tuning_curves.keys())
 
+def decode_2d(tuning_curves, group, ep, bin_size, xy, time_units="s", features=None):
+    warnings.warn(
+        "decode_2d is deprecated and will be removed in v1.0; use decode instead.",
```
sjvenditto:
```suggestion
        "decode_2d is deprecated and will be removed in a future version; use decode instead.",
```

doc/user_guide/07_decoding.md
```diff
@@ -23,18 +23,21 @@ import seaborn as sns
...
+
+<!-- #region -->
+Pynapple supports n-dimensional bayesian decoding. 
+The function returns the decoded feature as well as the probabilities for each timestamp. 
 
 
 :::{hint}
-Input to the bayesian decoding functions always include the tuning curves computed from [`nap.compute_1d_tuning_curves`](pynapple.process.tuning_curves.compute_1d_tuning_curves) or [`nap.compute_2d_tuning_curves`](pynapple.process.tuning_curves.compute_2d_tuning_curves).
+Input to the bayesian decoding functions always include the tuning curves computed using [`nap.compute_tuning_curves`](pynapple.process.tuning_curves.compute_tuning_curves).
 :::
```
sjvenditto, line 42: Similar to tuning curves, I would have a brief outline of the function signature and required inputs. We can also probably rename this section to "Bayesian Decoding" in preparation for the template matching/decoding

pynapple/process/decoding.py
wulfdewolf: Hmm I can do this but won't it be a bit weird? It'll be floats as column names?

wulfdewolf: I can do that, but isn't it weird to have floats as column names? It would be the bin centres that become the columns?

pynapple/process/decoding.py
wulfdewolf: No. I only put this in since someone might overwrite the attribute.

pynapple/process/decoding.py
wulfdewolf: Hmm. I prefer forcing `xarray` for everything (simplest), but if users want to pass dataframes I can concede.
The only use case I see for not wanting to pass the xarray would be if you compute the tuning curves yourself?

doc/user_guide/07_decoding.md
```diff
@@ -23,18 +23,27 @@ import seaborn as sns
 custom_params = {"axes.spines.right": False, "axes.spines.top": False}
 sns.set_theme(style="ticks", palette="colorblind", font_scale=1.5, rc=custom_params)
 ```
-Pynapple supports 1 dimensional and 2 dimensional bayesian decoding. The function returns the decoded feature as well as the probabilities for each timestamps. 
 
+Input to the decoding functions always includes:
+ - `tuning_curves`, computed using [`nap.compute_tuning_curves`](pynapple.process.tuning_curves.compute_tuning_curves).
+ - `data`, neural activity as a `TsGroup` (spikes), `TsdFrame` (e.g. smoothed rates), or dict of `Ts`/`Tsd`.
+ - `epochs`, to restrict decoding to certain intervals.
+ - `bin_size`, for when you pass spikes.
```
sjvenditto:
```suggestion
 - `bin_size`, the size of the bins in which to count the data (spikes only).
 - `time_units`, the units of `bin_size`, defaulting to seconds.
```

pynapple/process/decoding.py
```diff
@@ -19,194 +25,320 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
 
-        if tuning_curves.shape[1] != newgroup.shape[1]:
-            raise RuntimeError("Different shapes for tuning_curves and group")
+    Examples
+    --------
+    In the simplest case, we can decode a single feature (e.g., position) from a group of neurons:
+
+    >>> import pynapple as nap
+    >>> import numpy as np
+    >>> group = nap.TsGroup({i: nap.Ts(t=np.arange(0, 50) + 50 * i) for i in range(2)})
```
sjvenditto: Even though it still works as is, I would rename it to `data` to match the argument names (same with the following examples)

pynapple/process/decoding.py
```diff
@@ -19,194 +25,320 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
-            raise RuntimeError("Different indices for tuning curves and group keys")
-
-        count = newgroup.count(bin_size, ep, time_units)
-
-    elif type(group) is dict:
-        newgroup = nap.TsGroup(group, time_support=ep)
-        count = newgroup.count(bin_size, ep, time_units)
-
+    warnings.warn(
+        "decode_1d is deprecated and will be removed in a future version; use decode instead.",
```
sjvenditto:
```suggestion
        "decode_1d is deprecated and will be removed in a future version; use decode_bayes instead.",
```

also update for the other warnings

pynapple/process/decoding.py
```diff
@@ -19,194 +25,320 @@ def decode_1d(tuning_curves, group, ep, bin_size, time_units="s", feature=None):
...
-        if tuning_curves.shape[1] != len(newgroup):
-            raise RuntimeError("Different shapes for tuning_curves and group")
+        if not np.all(tuning_curves.coords["unit"] == np.array(list(data.keys()))):
+            raise ValueError("Different indices for tuning curves and data keys.")
 
-        if not np.all(tuning_curves.columns.values == np.array(newgroup.keys())):
-            raise RuntimeError("Different indices for tuning curves and group keys")
+        if isinstance(data, dict):
+            data = nap.TsGroup(data, time_support=epochs)
+        count = data.count(bin_size, epochs, time_units)
```
sjvenditto: Right now, a dict of Tsd objects will get caught in this section and therefore get counted. Are we assuming that the Tsds are still providing spike times, therefore making the values irrelevant? 

I think this is fine. I can't imagine a case where we would have a dictionary of Tsds with spike counts, because that should be a TsdFrame. If for some reason it's *not* a TsdFrame, then the length of each Tsd would need to be the same and concatenated into a TsdFrame, which sounds like obnoxious overhead over just assuming that it's spike times. Maybe we should document somewhere that dictionary objects are always interpreted to be spike times?

PR #485, opened 2025-08-02
Add `detect_oscillatory_events` method

jacobsimon:
### Summary
Adds a new method `detect_oscillatory_events` to the `filtering` module that allows for detection of ripple-like intervals in signals. Based on [this EEG processing example](https://github.com/PeyracheLab/pynacollada/blob/main/pynacollada/eeg_processing/eeg_processing.py). 

### Example
Screenshot 2025-08-02 at 2 38 15 PM

## Todo
* Will follow-up with a new user guide as discussed demonstrating usage on real data

### Open questions
* Still having trouble getting the sphinx docs to consistently compile - believe this is a local dev issue.
* We discussed using `apply_bandpass` instead of `filtfilt` but I think as written, both may be necessary unless we think it's ok to skip the smoothing step—or could use `Tsd.smooth` here?
* Does the testing approach make sense / seem adequate here?

### DIFF
--- a/pynapple/process/filtering.py
+++ b/pynapple/process/filtering.py
@@ -7,7 +7,7 @@ from numbers import Number
 
 import numpy as np
 import pandas as pd
-from scipy.signal import butter, sosfiltfilt, sosfreqz
+from scipy.signal import butter, filtfilt, sosfiltfilt, sosfreqz
 
 from .. import core as nap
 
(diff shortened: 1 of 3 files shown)


pynapple/process/filtering.py
```diff
@@ -510,3 +510,84 @@ def get_filter_frequency_response(
...
+def detect_oscillatory_events(
+    lfp, epoch, freq_band, thresh_band, duration_band, min_inter_duration, wsize=51
+):
+    """
+    Simple helper for detecting oscillatory events (e.g. ripples, spindles)
+
+    Parameters
+    ----------
+    lfp : Tsd
+        Should be a single channel raw lfp
```
gviejo: Not necessary raw lfp. I would just say
```suggestion
1-dimensional time series
```

pynapple/process/filtering.py
```diff
@@ -510,3 +510,84 @@ def get_filter_frequency_response(
...
+    freq_band : tuple
+        The (low, high) frequency to bandpass the signal
+    thresh_band : tuple
+        The (min, max) value for thresholding the normalized squared signal after filtering
+    duration_band : tuple
+        The (min, max) duration of an event in second
+    min_inter_duration : float
+        The minimum duration between two events otherwise they are merged (in seconds)
+    wsize : int, optional
+        The size of the window for digitial filtering
```
gviejo:
```suggestion
        The size of the window for digital filtering
```

pynapple/process/filtering.py
```diff
@@ -510,3 +510,84 @@ def get_filter_frequency_response(
...
+
+    Returns
+    -------
+    IntervalSet
+        The interval set of detected events with metadata containing
+        the power, amplitude, and peak_time
+    """
+    lfp = lfp.restrict(epoch)
+    frequency = lfp.rate
```
gviejo: This should be passed as a parameter 'fs' with default None.
If it's None, use `tsd.rate`. If it's a Number, use the passed `fs`.

pynapple/process/filtering.py
```diff
@@ -510,3 +510,84 @@ def get_filter_frequency_response(
...
+    wsize : int, optional
+        The size of the window for digitial filtering
+
+    Returns
+    -------
+    IntervalSet
+        The interval set of detected events with metadata containing
+        the power, amplitude, and peak_time
+    """
+    lfp = lfp.restrict(epoch)
```
gviejo:
```suggestion
    tsd = tsd.restrict(epoch)
```
It can be anything

pynapple/process/filtering.py
```diff
@@ -510,3 +510,84 @@ def get_filter_frequency_response(
...
+    # Merge if inter-oscillation period is too short
+    osc_ep = osc_ep.merge_close_intervals(min_inter_duration, time_units="s")
+
+    # Compute power, amplitude, and peak_time for each interval
+    powers = []
+    amplitudes = []
+    peak_times = []
+
+    for s, e in osc_ep.values:
+        seg = signal.restrict(nap.IntervalSet(s, e))
```
gviejo: Here you can be faster with : 
```suggestion
        seg = signal.get(s, e)
```

tests/test_filtering.py
```diff
@@ -523,3 +523,53 @@ def test_get_filter_frequency_response_error():
         ValueError, match="Unrecognized filter mode. Choose either 'butter' or 'sinc'"
     ):
         nap.get_filter_frequency_response(250, 1000, "lowpass", "a", 4, 0.02)
+
+
+@pytest.mark.parametrize(
+    "freq_band, thresh_band, start, end",
+    [
+        ((10, 30), (1, 10), 0, 2),
+        ((40, 60), (1, 10), 3, 5),
```
gviejo: There should be one more case which is the absence of detected events.

gviejo (changes requested):
Mostly minor comments but everything else is good.

pynapple/process/filtering.py
jacobsimon: Fixed in https://github.com/pynapple-org/pynapple/pull/485/commits/84fc65805dfe55fef626d6ed583fccce2701ed00

pynapple/process/filtering.py
jacobsimon: Changed to `data` in https://github.com/pynapple-org/pynapple/pull/485/commits/84fc65805dfe55fef626d6ed583fccce2701ed00

pynapple/process/filtering.py
jacobsimon: https://github.com/pynapple-org/pynapple/pull/485/commits/84fc65805dfe55fef626d6ed583fccce2701ed00

tests/test_filtering.py
jacobsimon: Added! I've noticed certain freq ranges seem to produce false positives though - going to keep testing a bit

gviejo:
Could you try to merge `dev`? The issue with the elife link should have been fixed within `dev`.

PR #495, opened 2025-10-06
489 refactor mutual information

wulfdewolf:
This combines:
- `compute_1d_mutual_info`
- `compute_2d_mutual_info`

into a generalised n-dimensional `compute_mutual_information`, which takes as only input the tuning curves as an xarray.DataArray (the new output of compute_tuning_curves) and returns an xarray.DataArray with dimensions "unit" (N) and "bits" (always 2, bits/spike and bits/sec).

Since occupancy is now an attribute of the tuning curves, `compute_mutual_information` uses that directly, and no other arguments need to be passed.

The old functions (and tests) are still available but they throw a DeprecationWarning.

### DIFF
--- a/pynapple/process/tuning_curves.py
+++ b/pynapple/process/tuning_curves.py
@@ -297,7 +297,9 @@ def compute_tuning_curves(
     keys = (
         data.keys()
         if isinstance(data, nap.TsGroup)
-        else data.columns if isinstance(data, nap.TsdFrame) else [0]
+        else data.columns
+        if isinstance(data, nap.TsdFrame)
+        else [0]
     )
     tcs = np.zeros([len(keys), *occupancy.shape])
     if isinstance(data, (nap.TsGroup, nap.Ts)):
@@ -500,7 +502,7 @@ def compute_discrete_tuning_curves(group, dict_ep):
 
 
 @_validate_tuning_inputs
-def compute_1d_mutual_info(tc, feature, ep=None, minmax=None, bitssec=False):
+def compute_1d_mutual_info_old(tc, feature, ep=None, minmax=None, bitssec=False):
     """
     Mutual information of a tuning curve computed from a 1-d feature.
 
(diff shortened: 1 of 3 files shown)


pynapple/process/tuning_curves.py
```diff
@@ -649,3 +649,145 @@ def compute_2d_mutual_info(dict_tc, features, ep=None, minmax=None, bitssec=Fals
...
         return SI
+
+
+def compute_mutual_information(tuning_curves):
+    """
+    Mutual information of an N-dimensional tuning curve.
+
+    See:
+
+    Skaggs, W. E., McNaughton, B. L., & Gothard, K. M. (1993).
```
BalzaniEdoardo: Add this in the "References" section of the numpy doc

doc/user_guide/06_tuning_curves.md
```diff
@@ -358,3 +349,17 @@ The output will be the mean firing rate of the neuron during this set of interva
 mean_fr = nap.compute_discrete_tuning_curves(tsgroup, dict_ep)
 print(mean_fr)
 ```
+
+# Mutual information
+Given a set of tuning curves, you can use [`compute_mutual_information`](pynapple.process.tuning_curves.compute_mutual_information) to compute the mutual information between the activity of the neurons and the features, no matter what dimension.
```
sjvenditto, line 354: Here I would also cite the paper that you cite in the method, I think it's a useful place to point users if they don't know what this is doing

tests/test_tuning_curves.py
```diff
@@ -712,6 +712,119 @@ def test_compute_discrete_tuning_curves(group, dict_ep):
...
+    ],
+)
+def test_compute_mutual_information_errors(tuning_curves, expectation):
+    with expectation:
+        nap.compute_mutual_information(tuning_curves)
+
+
+@pytest.mark.parametrize(
+    "n_units, n_features",
+    [(1, 1), (1, 2), (1, 3)],
```
sjvenditto: I would also add a couple cases with more than 1 unit

gviejo (approved):

PR #501, opened 2025-10-17
490 refactor discrete tuning curves

wulfdewolf:
This refactors `compute_discrete_tuning_curves` to return an `xarray,DataArray`, just like the new `compute_tuning_curves` function.

It is now also possible to pass a `TsdFrame`, in which case the mean over the epochs is computed.

There is a `return_pandas` argument for backwards compatibility.

Tests are updated.

### DIFF
--- a/pynapple/process/tuning_curves.py
+++ b/pynapple/process/tuning_curves.py
@@ -282,9 +282,9 @@ def compute_tuning_curves(
         return tcs
 
 
-def compute_discrete_tuning_curves(data, epochs_dict, return_pandas=False):
+def compute_response_per_epoch(data, epochs_dict, return_pandas=False):
     """
-    Compute discrete tuning curves using a dictionary of epochs.
+    Compute mean response per epoch, given a dictionary of epochs.
 
     Parameters
     ----------
@@ -310,7 +310,7 @@ def compute_discrete_tuning_curves(data, epochs_dict, return_pandas=False):
         ...     1: nap.Ts(np.arange(0, 100, 0.1)),
         ...     2: nap.Ts(np.arange(0, 100, 0.2))
         ... })
-        >>> tcs = nap.compute_discrete_tuning_curves(group, epochs_dict)
+        >>> tcs = nap.compute_response_per_epoch(group, epochs_dict)
         >>> tcs
         <xarray.DataArray (unit: 2, epochs: 2)> Size: 32B
         array([[10.03333333, 10.03333333],
@@ -322,7 +322,7 @@ def compute_discrete_tuning_curves(data, epochs_dict, return_pandas=False):
     You can also pass a TsdFrame (e.g. calcium imaging data), in that case the response is computed:
 
         >>> frame = nap.TsdFrame(d=np.random.rand(2000, 3), t=np.arange(0, 100, 0.05))
-        >>> tcs = nap.compute_discrete_tuning_curves(frame, epochs_dict)
+        >>> tcs = nap.compute_response_per_epoch(frame, epochs_dict)
         >>> tcs
         <xarray.DataArray (unit: 3, epochs: 2)> Size: 48B
         array([[0.50946668, 0.50897635],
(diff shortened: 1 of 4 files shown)


pynapple/process/tuning_curves.py
```diff
@@ -345,6 +278,281 @@ def compute_tuning_curves(
...
+        raise TypeError("data should be a TsdFrame, TsGroup, Ts, or Tsd.")
+
+    # check epochs_dict
+    if not isinstance(epochs_dict, dict) or not all(
+        isinstance(epoch, nap.IntervalSet) for epoch in epochs_dict.values()
+    ):
+        raise TypeError("epochs_dict should be a dictionary of IntervalSets.")
+
+    # check return_pandas
+    if not isinstance(return_pandas, bool):
```
sjvenditto: do we also want to allow integer values of `0` and `1`? right now that breaks this (I see there's the same restriction in `compute_tuning_curves`)

pynapple/process/tuning_curves.py
```diff
@@ -345,6 +278,281 @@ def compute_tuning_curves(
...
+    # tuning curves
+    keys = (
+        data.keys()
+        if isinstance(data, nap.TsGroup)
+        else data.columns if isinstance(data, nap.TsdFrame) else [0]
+    )
+    tcs = np.empty((len(keys), len(epochs_dict)))
+    if isinstance(data, (nap.TsGroup, nap.Ts)):
+        # SPIKES
+        if isinstance(data, nap.Ts):
```
sjvenditto, line 368: you could define the `Ts` as a `TsGroup` and then do list comprehension with class methods like
```
tcs = np.stack([group.restrict(epoch).count().values.sum(axis=0) / epoch.tot_length("s") for epoch in epochs_dict.values()])
```
theoretically we shouldn't need `.values` per se, but it looks like there's a bug when summing along axis 0 for a 2x2 `TsdFrame`

pynapple/process/tuning_curves.py
```diff
@@ -345,6 +278,281 @@ def compute_tuning_curves(
...
+            data = {0: data}
+        for epoch_idx, epoch in enumerate(epochs_dict.values()):
+            for unit_idx, unit_label in enumerate(keys):
+                tcs[unit_idx, epoch_idx] = float(len(data[unit_label].restrict(epoch)))
+            tcs[:, epoch_idx] = tcs[:, epoch_idx] / epoch.tot_length("s")
+    else:
+        # RATES
+        if isinstance(data, nap.Tsd):
+            data = np.expand_dims(data.values, -1)
+        for epoch_idx, epoch in enumerate(epochs_dict.values()):
```
sjvenditto: similarly:
```
tcs = np.stack([frame.restrict(epoch).values.mean(axis=0) for epoch in epochs_dict.values()])
```

but also, wouldn't want the average response per second? like:
```
tcs = np.stack([frame.restrict(epoch).values.sum(axis=0) / epoch.tot_length() for epoch in epochs_dict.values()])
```

tests/test_tuning_curves.py
```diff
@@ -611,107 +611,381 @@ def test_compute_tuning_curves_type_errors(data, features, kwargs, expectation):
...
+            get_group_n(2).count(1.0),
+            {"0": nap.IntervalSet(0, 50)},
+            {},
+            xr.DataArray(
+                [[10.0], [1.0]],
+                dims=["unit", "epochs"],
+                coords={"unit": [1, 2], "epochs": ["0"]},
+            ),
+        ),
+        # two rate units, multiple epochs
```
sjvenditto, line 822: we also want to test when there are multiple intervals in each "stimulus" group, and when stimulus intervals overlap

pynapple/process/tuning_curves.py
```diff
@@ -345,6 +278,281 @@ def compute_tuning_curves(
         return tcs
 
 
+def compute_discrete_tuning_curves(data, epochs_dict, return_pandas=False):
+    """
+    Compute discrete tuning curves using a dictionary of epochs.
```
sjvenditto: Maybe we add a bit saying that intervals across different epochs can overlap (but not within epochs)

pynapple/process/tuning_curves.py
wulfdewolf: List comprehension: _**yes**_
Average response per second: not sure, I'm imagining calcium imaging activity as input. 
That has arbitrary unit? I think we'd just want the mean for that?
That's what we're doing in the normal tuning curves: the output is in the same unit as the input.

gviejo (approved):

PR #503, opened 2025-10-17
Neo io

BalzaniEdoardo:
Start off a neo-io support.

### DIFF
--- a/pynapple/io/interface_neo.py
+++ b/pynapple/io/interface_neo.py
@@ -1119,24 +1119,14 @@ class EphysReader(UserDict):
         sampling_rate = float(proxy.sampling_rate.rescale("Hz").magnitude)
         n_samples = proxy.shape[0]
 
-        def _loader(
-            _fp=file_path,
-            _dt=dtype,
-            _bs=buf_shape,
-            _cs=col_slice,
-            _t0=t_start,
-            _sr=sampling_rate,
-            _ns=n_samples,
-            _nt=nap_type,
-            _meta=metadata,
-        ):
-            fp = np.memmap(_fp, _dt, "r", shape=_bs)
-            data = fp if _cs is None else fp[:, _cs]
-            timestamps = _t0 + np.arange(_ns) / _sr
+        def _loader():
+            fp = np.memmap(file_path, dtype, "r", shape=buf_shape)
+            data = fp if col_slice is None else fp[:, col_slice]
+            timestamps = t_start + np.arange(n_samples) / sampling_rate
             kwargs = {}
-            if _meta is not None:
-                kwargs["metadata"] = _meta
-            return _nt(t=timestamps, d=data, load_array=False, **kwargs)
+            if metadata is not None:
+                kwargs["metadata"] = metadata
+            return nap_type(t=timestamps, d=data, load_array=False, **kwargs)
 
         return {
             "type": nap_type.__name__,


doc/user_guide/02_input_output.md
```diff
@@ -13,13 +13,16 @@ kernelspec:
 
 # Input-output & lazy-loading
 
-Pynapple provides loaders for [NWB format](https://pynwb.readthedocs.io/en/stable/index.html#). 
+Pynapple provides multiple ways to load data. The two main formats are NWB and NPZ. In addition, raw LFP data can be loaded through the NEO library.
```
wulfdewolf: I wouldn't say LFP, maybe:
In addition, raw data can be loaded using [NEO](https://neo.readthedocs.io/en/latest/).

doc/user_guide/02_input_output.md
```diff
@@ -13,13 +13,16 @@ kernelspec:
 
 # Input-output & lazy-loading
 
-Pynapple provides loaders for [NWB format](https://pynwb.readthedocs.io/en/stable/index.html#). 
+Pynapple provides multiple ways to load data. The two main formats are NWB and NPZ. In addition, raw LFP data can be loaded through the NEO library. 
 
 Each pynapple objects can be saved as a [`npz`](https://numpy.org/devdocs/reference/generated/numpy.savez.html) with a special structure and loaded as a `npz`.
```
wulfdewolf: Each Pynapple object can be saved and loaded as a [`.npz`](https://numpy.org/devdocs/reference/generated/numpy.savez.html) with a Pynapple-specific structure.

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
 print(type(z.d))
 ```
 
+## LFP loading & NEO compatibility
+
+Raw LFP data can be loaded with pynapple through the NEO library.
```
wulfdewolf: I guess we need to decide if we zoom in on LFP or not.
Many of the modern probes like the Neuropixels don't explicitly record LFP, so why not just say raw data? And then LFP can be a subset.

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+Internally, pynapple uses the NEO raw IO classes to read the data and convert them to one of the pynapple time series object.
```
wulfdewolf: objectS

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+This is done through the class [`nap.EphysReader`](pynapple.io.interface_neo.EphysReader).
```
wulfdewolf: the [`nap.EphysReader`](pynapple.io.interface_neo.EphysReader) class.

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+
+See here the [list of supported formats](https://neo.readthedocs.io/en/stable/rawiolist.html) of python-neo.
```
wulfdewolf, line 147: The list of formats supported by `python-neo` can be found [here](https://neo.readthedocs.io/en/stable/rawiolist.html).

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+
+The minimal example to load a dataset is to instantiate the `EphysReader` class with the path to the data file.
+
+```
+import pynapple as nap
+data = nap.EphysReader("path_to_your_file")
+```
+
+Below are shown a few examples of loading LFP and spikes.
```
wulfdewolf: Let us look at a couple of examples where we load LFP and spike data.

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+
+### Neuroscope / Binary file
+
+If you have a session recorded as binary files with Neuroscope, you can load it with the `EphysReader` class or directly 
+with the function [`nap.load_eeg`](pynapple.io.misc.load_eeg).
```
wulfdewolf: EEG stand for electroencephalography, imo it should have a different name

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+    5e-05        0.262146   0.205078   0.325928   0.163574   0.235901  ...
+    0.0001       0.263062   0.221863   0.325317   0.181885   0.266113  ...
+    ...                                                                ...
+    1199.99585  -0.181885  -0.186157  -0.283508  -0.15686   -0.209961  ...
+    1199.9959   -0.098877  -0.209961  -0.291138  -0.213623  -0.209656  ...
+    1199.99595  -0.109863  -0.141296  -0.257874  -0.188293  -0.163879  ...
+    dtype: int16, shape: (23999920, 16)
+```
+
+A more direct way to load binary file is to use the function `nap.load_eeg` that will directly return a `TsdFrame` object.
```
wulfdewolf: to use the `nap.load_eeg` function, which will return `TsdFrame` object directly.

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+          │  ┗ 📄 timestamps.npy
+          ┗ 📂 Neuropix-PXI-103.ProbeA
+             ┗ 📂 TTL
+                ┣ 📄 full_words.npy
+                ┣ 📄 sample_numbers.npy
+                ┣ 📄 states.npy
+                ┗ 📄 timestamps.npy
+```
+
+You can load the LFP data with the `EphysReader` class with :
```
wulfdewolf: This is a recording from a Neuropixel 2.0 probe, which does not have dedicated LFP channels. Instead, we can visualize the raw data directly, using `EphysReader` as follows:

doc/user_guide/02_input_output.md
```diff
@@ -133,6 +138,177 @@ z = data['z']
...
+    8.044433333   -132.015  -90.87   -105.3    -187.59   -122.85   ...
+    ...                                                            ...
+    68.606033333    33.54    56.55     50.31    -30.615     6.045  ...
+    68.606066667    35.88    56.55     45.045   -35.88     -1.56   ...
+    68.6061         32.76    61.815    51.87    -33.54      7.605  ...
+    dtype: int16, shape: (1816853, 384)
+    
+```
+
+### Plexon file
```
wulfdewolf, line 281: If you have Plexon files, you can use `EphysReader` in the exact same way:

doc/user_guide/02_input_output.md
wulfdewolf: I feel like we should mention that things are lazy-loaded?

wulfdewolf (commented):
Looks good to me!

My comments are mostly for the user guide: small language comments and raising the question whether we should focus only on LFP or just saying we can load any "raw" data, which I think is better.

I managed to run it locally and it runs very smoothly, even when the data is not even downloaded locally but on some network-mounted drive.

The tests look comprehensive. I'm imagining once people start using this things might pop-up, and we might have to add tests to catch them, but I think it wouldn't be very practical trying to predict them now.

doc/user_guide/02_input_output.md
gviejo: Agreed

doc/user_guide/02_input_output.md
gviejo: Agreed

doc/user_guide/02_input_output.md
gviejo: True it's a very early function. in the neurosuite format, downsampled lfp sometimes have the extension .eeg. Maybe it is the right PR to rename it `nap.load_binary_file`?

pynapple/io/interface_neo.py
```diff
@@ -0,0 +1,1403 @@
...
+    Examples
+    --------
+    >>> interface = NeoSignalInterface(signal, block, time_support, sig_num=0)
+    >>> # Use as array-like for lazy loading
+    >>> tsd = nap.Tsd(t=interface.times, d=interface, load_array=False)
+    >>> # Data is only loaded when accessed
+    >>> chunk = tsd[0:1000]  # Loads only first 1000 samples
+    """
+
+    def __init__(self, signal, block, time_support=None, sig_num=0):
```
BalzaniEdoardo, line 567: I think this init does too much, you should probably split it in a bunch of function calls that do stuff like crating the offset array etc. In this way it will be easier to unit test these too

pynapple/io/interface_neo.py
```diff
@@ -0,0 +1,1403 @@
...
+        dtype = np.dtype(bd["dtype"])
+        buf_shape = tuple(bd["shape"])
+        col_slice = reader._stream_buffer_slice[stream_id]
+
+        # Pre-compute timestamps from the proxy metadata
+        t_start = _rescale_to_seconds(proxy.t_start)
+        sampling_rate = float(proxy.sampling_rate.rescale("Hz").magnitude)
+        n_samples = proxy.shape[0]
+
+        def _loader(
```
BalzaniEdoardo: do you ever change the input to this loader function? otherwise why not doing this

```python
def _loader():
    # your loader logic using directly the variable defined in the outer scope
    fp = np.memmap(file_path, dtype, "r", shape=buf_shape)
    # ....
```

BalzaniEdoardo (commented):
Looking very neat

pynapple/io/interface_neo.py
gviejo: Can you detail more the suggestion here?

BalzaniEdoardo: Since `_make_mmap_entry` is a method and not called inside a loop, every invocation creates a fresh local scope — `file_path`, `dtype`, `buf_shape`, etc. are local variables that won't be reassigned after `_loader` is defined. So the closure can reference them directly without risk of capturing a stale value. You can simplify to:

```python
def _loader():
    fp = np.memmap(file_path, dtype, "r", shape=buf_shape)
    data = fp if col_slice is None else fp[:, col_slice]
    timestamps = t_start + np.arange(n_samples) / sampling_rate
    kwargs = {}
    if metadata is not None:
        kwargs["metadata"] = metadata
    return nap_type(t=timestamps, d=data, load_array=False, **kwargs)
 ```
 
 The default-argument pattern (`_fp=file_path`, etc.) is a guard against late binding, which only matters when closures are created inside a loop over a mutating variable. It's not needed here.

gviejo: Thanks updating it

Issue #520, opened 2025-11-03
KeyError: 'data is not a file in the archive' when loading Tsg containing Tsd from .npz file


danielsvedberg:
https://github.com/pynapple-org/pynapple/blob/c63c36b5ee3890fb810bf4b758ab415e4d61915a/pynapple/core/ts_group.py#L1623C11-L1623C12

Hi,

I've had this recurring issue where if I create and save Tsg containing a Tsd, then loading the file will throw the error:
`KeyError: 'data is not a file in the archive'`
I do not get this error when loading a Tsg composed only of Ts objects.

running debug mode on the error pulls up the following snippet as the snag point:
> c:\users\user\miniconda3\envs\si_env\lib\site-packages\pynapple\core\ts_group.py(1623)_from_npz_reader()
   1621 
   1622         if has_data:
-> 1623             data = file["data"]
   1624 
   1625         if "keys" in file.keys():

I'm guessing this is because line 1619 in ts_group.py conditions `has_data` on `file["d"]` but then queries `file["data"]` instead.

gviejo:
Thanks good catch.

PR #521, opened 2025-11-03
fixed incorrect "data" field reference during Tsg loading

danielsvedberg:
In reference to #520

Just implemented the small fix and tested successfully

### DIFF
--- a/tests/test_npz_file.py
+++ b/tests/test_npz_file.py
@@ -55,14 +55,18 @@ data = {
     "tsdgroup": nap.TsGroup(
         {
             0: nap.Tsd(t=np.arange(0, 200), d=np.random.rand(200)),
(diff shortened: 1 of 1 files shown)


gviejo:
Thanks for the PR. Would it be possible to add a test case in [tests/test_npz_file.py](https://github.com/pynapple-org/pynapple/blob/c63c36b5ee3890fb810bf4b758ab415e4d61915a/tests/test_npz_file.py#L28)?

danielsvedberg:
Just added some unit tests for 'tsdgroup'. The more streamlined approach would be to edit the tsgroup called by test_load_tsgroup so that it contains a mix of Ts and Tsd. I went for the approach of leaving test_load_tsgroup intact and creating a separate tsdgroup (containing a mix of ts and tsd) and a test_load_tsdgroup function, to isolate the mixed tsgroup case into a separate test. 

I'm new to contributing so I apologize in advance if I did something wrong!

gviejo (approved):

Issue #538, opened 2026-01-13
Add sparsity score (like mutual information)


pas-calc:
Similar to information score (mutual information), we could add sparsity score.

https://pynapple.org/generated/pynapple.process.tuning_curves.html#pynapple.process.tuning_curves.compute_mutual_information
https://pynapple.org/_modules/pynapple/process/tuning_curves.html#compute_mutual_information

Formula: Sparsity = <λ>²/<λ²> where λ is the mean firing rate per bin

So in python it would be
```
# occupancy is already a probability distribution (normalized to have sum 1)
fr = np.nansum(fx * occupancy, axis=axes) # expectation value of firing rate per bin (from tuning curves)
fr2 = np.nansum(fx**2 * occupancy, axis=axes) # expectation value of firing rate squared
Sparsity = fr**2/fr2 # formula for variance
```

Paper:

Skaggs WE, McNaughton BL, Wilson MA, Barnes CA. Theta phase precession in hippocampal neuronal populations and the compression of temporal sequences. Hippocampus. 1996;6(2):149-72. doi: 10.1002/(SICI)1098-1063(1996)6:2<149::AID-HIPO6>3.0.CO;2-K. PMID: 8797016.

Chapter: "Spatial Firing Measures"

PR #539, opened 2026-01-13
add sparsity score to tuning curves

pas-calc:
see https://github.com/pynapple-org/pynapple/issues/538

### DIFF
--- a/pynapple/process/__init__.py
+++ b/pynapple/process/__init__.py
@@ -37,6 +37,7 @@ from .tuning_curves import (
     compute_2d_tuning_curves_continuous,
     compute_discrete_tuning_curves,
     compute_mutual_information,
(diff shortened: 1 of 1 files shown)


pas-calc:
> I like the idea of this! The sparsity score is indeed nice to have, and easily computed from the tuning curves. The function you wrote seems good.
> 
> For us to merge this though we will need:
> 
>     * comprehensive testing
> 
>     * docstring examples
> 
> 
> As we have for the mutual information function.
> 
> The fasted way will be if you write these! I can also do it, but have other priorities for the moment, so that route will take longer.

Yes I will have a go on that

gviejo:
Hello you can lint the project with `black pynapple` after installing the right version `pip install black==26.1.0`.

Issue #537, opened 2026-01-13
pynapple.process.randomize : shift_timestamps , correct rolling behavior


pas-calc:
The current definition (from `_shift_ts`) is
`shifted_timestamps = (ts.times() + shift) % ts.end_time() + ts.start_time()`

https://pynapple.org/generated/pynapple.process.randomize.html#pynapple.process.randomize.shift_timestamps
https://pynapple.org/_modules/pynapple/process/randomize.html#shift_timestamps

However I am wondering if this is correct?

I am suggesting this definition
`shift_x = (x - start + shift) % duration + start`

```
# example
x=np.array([25,27,33.3,34.5])
start=x[0]
end=x[-1]
duration=end-start
shift = 1

>>> x
array([25. , 27. , 33.3, 34.5])
>>> (x - start + shift) % duration + start
array([26. , 28. , 34.3, 26. ])

```

But the current implementation yields:
```
>>> (x + shift) % end + start
array([51. , 53. , 59.3, 26. ])
```
These timestamps are not rolled from the original sequence but have values outside the range.

Related https://github.com/pynapple-org/pynapple/issues/528

PR #540, opened 2026-01-13
537 fix timestamps shift

wulfdewolf:
This addresses [#537](https://github.com/pynapple-org/pynapple/issues/537) by correctly implementing timestamp shifting.
It adds comprehensive testing to avoid the issue in the future.
It adds docstring examples to `shift_timestamps`.
It adds a warning when a `TsGroup` is passed that contains a `Tsd` in any entry, the warning states that we will treat it as a `Ts`.

### DIFF
--- a/pynapple/process/randomize.py
+++ b/pynapple/process/randomize.py
@@ -31,10 +31,11 @@ def shift_timestamps(data, min_shift=0.0, max_shift=None, mode="drop"):
         minimum shift (default: 0)
     max_shift : float, optional
         maximum shift, (default: length of time support)
-    mode : {'drop', 'wrap'}, optional
+    mode : ``'drop'`` or ``'wrap'``, optional
         How to handle timestamps that fall outside the time support after shifting.
-        - 'drop' (default): drop those timestamps
-        - 'wrap': circularly wrap timestamps within the time support
+
+        * ``'drop'``: (default): drop those timestamps
+        * ``'wrap'``: circularly wrap timestamps within the time support
 
     Returns
     -------
--- a/doc/user_guide/09_randomization.md
+++ b/doc/user_guide/09_randomization.md
@@ -28,7 +28,8 @@ Different methods preserve or destroy different features of the data.
 
 ## Shift timestamps
 
-[`shift_timestamps`](pynapple.process.randomize.shift_timestamps) shifts all the timestamps in a `Ts` object by the same random amount, wrapping the end of the time support to its beginning. This randomization preserves the temporal structure in the data but destroys the temporal relationships with other quantities (e.g. behavioural data).
+[`shift_timestamps`](pynapple.process.randomize.shift_timestamps) shifts all the timestamps in a `Ts` object by the same random amount.
+This randomization preserves the temporal structure in the data but destroys the temporal relationships with other quantities (e.g. behavioural data).
 When applied on a `TsGroup` object, each series in the group is shifted independently.
 
 


pynapple/process/randomize.py
```diff
@@ -2,41 +2,111 @@
...
         The randomly shifted timestamps
+
+    Examples
+    --------
+
+    As an example, consider the case where we set a fixed shift:
+
+        >>> import pynapple as nap
+        >>> ts = nap.Ts([25, 27, 33.3, 34.5])
+        >>> shifted_ts = nap.shift_timestamps(ts, min_shift=1, max_shift=1)
```
gviejo: I think the example is wrong here. What happen when min_shift = max_shift?

pynapple/process/randomize.py
```diff
@@ -2,41 +2,111 @@
...
+        shape: 4
+
+    If the input has a time support, it will be respected:
+
+        >>> epochs = nap.IntervalSet(start=0, end=34.5)
+        >>> ts = nap.Ts([25, 27, 33.3, 34.5], time_support=epochs)
+        >>> shifted_ts = nap.shift_timestamps(ts, min_shift=1, max_shift=1)
+        >>> shifted_ts
+        Time (s)
+        1.0
```
gviejo: Here I don't understand where the 1.0 comes from?

pynapple/process/randomize.py
gviejo: Ok I would say it's a very weird case. It's not obvious that the shift will be of 1 every time.

gviejo: Which is what you said above ok.

gviejo:
```suggestion
        >>> shifted_ts = nap.shift_timestamps(ts, min_shift=1, max_shift=1) # Here shift = 1.0 second
```

pynapple/process/randomize.py
gviejo: Ok it rolls within the time support but I though we said it would be dropped?

pynapple/process/randomize.py
```diff
@@ -2,41 +2,111 @@
...
+        34.3
+        shape: 4
+
+    If the time support has multiple epochs, the timepoints that fall outside of these
+    epochs after shifting will be dropped:
+
+        >>> epochs = nap.IntervalSet(start=[25, 30], end=[27, 34.5])
+        >>> ts = nap.Ts([25, 27, 33.3, 34.5], time_support=epochs)
+        >>> shifted_ts = nap.shift_timestamps(ts, min_shift=1, max_shift=1)
+        >>> shifted_ts
```
gviejo, line 89: Ok that's confusing with above. I don't understand why the previous one is not dropped either?

pynapple/process/randomize.py
wulfdewolf: When `min_shift == max_shift` then:
```
shift = np.random.uniform(min_shift, max_shift)
```
gives `1`.
I'm happy to use an example with a random shift, it's just harder to explain...

pynapple/process/randomize.py
wulfdewolf: The `time_support` was set to `[0, 34.5]` so it wraps differently. As an example I thought, but happy to remove it.

pas-calc: > Here I don't understand where the 1.0 comes from?

The 1 is 0+1 (1 is the shift) where 0=34.5 (the last event is temporally equivalent to 0 because of the wrapping)

25 +1 = 26
27 +1 = 28
33.3 +1 = 34.3
34.5 +1 = 35.5 mod 34.5 = 1

So the new sorted time series is [1,26,28,34.3]

Applying modulo on the other events won't affect anything since they do not wrap (<34.5)

PR #554, opened 2026-02-05
Correct deprecation warnings in tuning_curves.py

bendichter:


wulfdewolf:
I have addressed this in a separate PR, thanks for pointing it out!

Issue #615, opened 2026-07-14
Bug: IntervalSet drops metadata when start/end contain NaN


AbstractGeek:
Thanks for this fantastic library! 

Pynapple appears to mishandle `IntervalSet` construction when the input dataframe contains `NaN` values in the `start` or `end` columns. Some DANDI datasets include `NaN` values in interval start/stop times and the bug can be triggered by those data. The minimal code below describes this bug.

```python
import numpy as np
import pandas as pd
import pynapple as nap


df = pd.DataFrame(
    {
        "start": [0.0, np.nan],
        "end": [1.0, 2.0],
        "label": ["a", "b"],
    }
)

intervals = nap.IntervalSet(df)
print(intervals)
print(intervals.metadata)
```

Observed behavior:
```text
<string>:9: UserWarning: start is not sorted. Cannot add metadata to unsorted start times. Sorting it and dropping metadata.
```

Expected behavior:
- The `IntervalSet` should preserve metadata columns even when `start`/`end` contain `NaN` values.
- If `NaN` values are present, the code should display a user warning, automatically remove the invalid rows, and provide a valid `IntervalSet` with the remaining metadata intact.

Example dandi set with NaN in stop times:  dandiset: '000620'; path: 'sub-Elgar/sub-Elgar_ses-2022-08-19_behavior+task.nwb'

wulfdewolf:
Hi @AbstractGeek , thanks for submitting this issue and minimal reproducible code!
We agree with your solution, I'm working on implementing it right now.

gviejo:
Hi Dinesh. Fix should appear in the next release. If you want to test it right now, you can checkout to the dev branch: `git checkout dev`

PR #616, opened 2026-07-14
Drop intervals with `np.nan` in times

wulfdewolf:
### Summary
This short PR adds checks in the `__init__` function of `IntervalSet` to find NaNs in the start and end times, dropping them and their corresponding metadata.
A warning is thrown indicating how many intervals are dropped.

### Tests
I added thorough testing to verify that the right entries are dropped, that it works no matter the input type, and that the warning is printed.

### Related
#615


gviejo (commented):
Looks good. merging it

PR #617, opened 2026-07-15
First comit for movement integration

gviejo:
>>> import pynapple as nap
    >>> data = nap.from_movement(ds)
    >>> data
    test.h5  (poses | 100 frames | 10.0 fps)
    ╭──────────────┬────────────────┬──────────┬──────────────────────────────────────╮
    │ Individual   │ Variable       │ Type     │ Columns / Shape                      │
    ├──────────────┼────────────────┼──────────┼──────────────────────────────────────┤
    │ mouse1       │ position       │ TsdFrame │ nose_x, nose_y, left_ear_x, ...  (6) │
    │ mouse1       │ confidence     │ TsdFrame │ nose, left_ear, right_ear  (3)       │
    ╰──────────────┴────────────────┴──────────┴──────────────────────────────────────╯
    >>> data['position']
    Time (s)    nose_x    nose_y    ...

This pull request adds a new interface for importing and working with datasets from the `movement` package in Pynapple. It introduces a robust conversion utility (`from_movement`) and a dict-like container (`MovementDataset`) for handling pose and bounding box data, with support for multiple individuals and lazy loading of time series. Additionally, comprehensive tests and a minimal user example are included.

**Movement package integration:**

* Added `from_movement` function and `MovementDataset` class to `pynapple.io.movement`, enabling conversion of movement `xarray.Dataset` objects (poses/bboxes) into Pynapple time series objects, with validation, multi-individual support, and a rich preview table.
* Registered `from_movement` in the Pynapple IO API (`pynapple/io/__init__.py`).

**Testing and documentation:**

* Added comprehensive tests for the new interface, covering validation, single/multiple individual datasets, lazy loading, and preview table output (`tests/test_movement.py`).
* Provided a minimal example script demonstrating the workflow from loading movement data to conversion and inspection in Pynapple (`movement_example.py`).

### DIFF
--- a/movement_example.py
+++ b/movement_example.py
@@ -1,0 +1,53 @@
+"""
+Movement → Pynapple minimal example
+Based on: https://movement.neuroinformatics.dev/latest/examples/scale.html
+"""
+
+import warnings
+from movement import sample_data
+from movement.filtering import filter_by_confidence, interpolate_over_time, median_filter
+from movement.transforms import scale
+import pynapple as nap
+
+warnings.filterwarnings("ignore")
+
+# ---------------------------------------------------------------------------
+# Load and pre-process
+# ---------------------------------------------------------------------------
+
+ds = sample_data.fetch_dataset("DLC_single-mouse_DBTravelator_2D.predictions.h5")
+
+kp_dim  = "keypoint"   if "keypoint"   in ds.coords else "keypoints"
+ind_dim = "individual" if "individual" in ds.coords else "individuals"
+
+landmark_keypoints = [
+    "Door", "StartPlatL", "StartPlatR", "StepL", "StepR", "TransitionL", "TransitionR",
+]
+
+ds_mouse = ds.sel(
+    {
+        kp_dim:  ~ds[kp_dim].isin(landmark_keypoints),
+        ind_dim: ds[ind_dim].values[0],
+    }
+)
+
+ds_mouse["position"] = filter_by_confidence(ds_mouse.position, ds_mouse.confidence, threshold=0.9)
+ds_mouse["position"] = interpolate_over_time(ds_mouse.position, max_gap=40)
+ds_mouse["position"] = median_filter(ds_mouse.position, window=6, min_periods=2)
+
+# Scale pixels → cm (1 cm grid ≈ 29 px) and flip y-axis
+ds_mouse["position"] = scale(ds_mouse["position"], factor=1.0 / 29.0, space_unit="cm")
+y = ds_mouse["position"].sel(space="y")
+ds_mouse["position"].loc[dict(space="y")] = y.max() - y
+
+# ---------------------------------------------------------------------------
+# Convert to pynapple
+# ---------------------------------------------------------------------------
+
+data = nap.from_movement(ds_mouse)
+print(data)
+
+position   = data["position"]    # TsdFrame — columns: Nose_x, Nose_y, EarL_x, ...
+confidence = data["confidence"]  # TsdFrame — columns: Nose, EarL, ...
+
+print(position)
--- a/pynapple/io/__init__.py
+++ b/pynapple/io/__init__.py
@@ -1,4 +1,5 @@
 from .folder import Folder
+from .movement import from_movement
 from .interface_neo import EphysReader, NeoSignalInterface
 from .interface_npz import NPZFile
 from .interface_nwb import NWBFile
--- a/pynapple/io/movement.py
+++ b/pynapple/io/movement.py
@@ -1,0 +1,314 @@
+"""
+Pynapple interface for the movement package.
+Converts movement xarray.Dataset objects to pynapple time series.
+"""
+
+from collections import UserDict
+
+import numpy as np
+from tabulate import tabulate
+
+from .. import core as nap
+
+
+def _check_xarray():
+    try:
+        import xarray  # noqa: F401
+    except ImportError:
+        raise ImportError(
+            "The 'xarray' package is required to use from_movement. "
+            "Install it with: pip install xarray"
+        )
+
+
+def _dim(ds, *candidates):
+    """Return the first candidate that exists as a coordinate in ds."""
+    for name in candidates:
+        if name in ds.coords:
+            return name
+    raise KeyError(
+        f"None of {candidates} found in dataset coordinates: {list(ds.coords)}"
+    )
+
+
+def _ind_values(ds):
+    """Return the list of individual names, handling scalar (dropped-dim) case."""
+    ind_dim = _dim(ds, "individual", "individuals")
+    val = ds.coords[ind_dim].values.tolist()
+    # After ds.sel(individuals="name") the coord becomes a 0-d scalar string
+    if isinstance(val, str):
+        return [val]
+    return val
+
+
+def _validate_movement_dataset(ds):
+    """Validate that ds is a supported movement xarray.Dataset."""
+    _check_xarray()
+    import xarray as xr
+
+    if not isinstance(ds, xr.Dataset):
+        raise TypeError(f"Expected an xarray.Dataset, got {type(ds).__name__}.")
+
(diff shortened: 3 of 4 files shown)


niksirbi:
Great to see this @gviejo !

Let me know if you need any input from our (`movement` devs) side, of if we can help with reviewing this functionality.

gviejo:
Hi @niksirbi I wrote the comments I had when testing this in this PR (https://github.com/pynapple-org/pynaviz/pull/118). 

There is an example file in this PR doing skeleton visualization (https://github.com/pynapple-org/pynaviz/pull/118/changes#diff-4d8015ff09e96164b44f420abb5be9e817714de3a81f0975a40edc97fda97c4d)

I think the biggest comment I had was about storing the skeleton informations. Right now in the example, I manually pass the skeleton information. It would be ideal to retrieve it automatically.

gviejo:
Btw the example will generate this visualization. 
image

It would be nice to have an example that mixes tracking data and physiological data recorded simultaneously.

niksirbi:
That looks so cool! Let me have a detailed look at your comments and your example and I'll get back to you. Thanks for getting the ball rolling on this!

niksirbi:
Hey @gviejo, me and @lochhh did some exploration of what it would take to achieve a meaningful `movement`-to-`pynapple` integration, inspired by your work.

Here are our many random thoughts on this, and we'd be keen to get your opinion on them.

## The NWB route

There is already an existing route for going from `movement` datasets to pynapple objects, going through NWB first.

This looks something like this (using one of our sample datasets to illustrate):

```python
>>> import numpy as np
>>> import pynapple as nap
>>> from movement import sample_data
>>> from movement.io import save_poses
>>>
>>> ds = sample_data.fetch_dataset("DLC_single-wasp.predictions.h5")
>>> nwb_file = save_poses.to_nwb_file(ds) # returns a pynwb.file.NWBFile
>>> nap_file = nap.NWBFile(nwb_file)
>>> print(nap_file)
-
┍━━━━━━━━━┯━━━━━━━━━━┑
│ Keys    │ Type     │
┝━━━━━━━━━┿━━━━━━━━━━┥
│ stinger │ TsdFrame │
│ head    │ TsdFrame │
┕━━━━━━━━━┷━━━━━━━━━━┙
>>> print(nap_file["head"])
Time (s)          x        y
----------  -------  -------
0.0         1086.24  421.501
0.025       1086.02  421.544
0.05        1085.54  421.807
0.075       1085.08  422.59
0.1         1085.19  422.488
0.125       1084.93  422.536
0.15        1084.85  422.781
...
26.95          0       0
26.975         0       0
27.0           0       0
27.025         0       0
27.05          0       0
27.075         0       0
27.1           0       0
dtype: float64, shape: (1085, 2)
```

However, there are some drawbacks of this approach:
- The confidence scores are not recovered
- Keypoints appear as separate TsdFrames. This could get awkward if there are many of them.
- It only works for single animal data (because NWB forces that)
- It doesn't work for bounding boxes datasets
- If there were additional variables in the `movement` ds (e.g. `head_direction`), they would not natively survive this trip through NWB.

We use the https://github.com/rly/ndx-pose extension to read/write pose estimation time series to/from NWB files, but as far as I can tell `pynapple` doesn't (yet) 'understand' `ndx-pose`. We could probably improve the situation by making a few tweaks in how both packages (`movement` and `pynapple`) read/write NWB files, for example:

- `pynapple` could add cases for loading `ndx-pose` objects [here](https://github.com/pynapple-org/pynapple/blob/359fe8d5167f4f7ca544a23532e30ba462b50313/pynapple/io/interface_nwb.py#L60)
- `movement` could probably make better use of [pynwb.behavior](https://pynwb.readthedocs.io/en/latest/pynwb.behavior.html#module-pynwb.behavior) to also store additional (non-position) behavioural variables as `SpatialSeries`.

## The direct route

This is basically what you've implemented here, going directly from an `xarray` object to a `pynapple` object, entirely bypassing NWB. Seems to work nicely.

My main question on this is: do you think this conversion functionality best belongs in `pynapple` or in `movement`? For now I lean towards having a `movement.save_poses.to_pynapple()` exporter (and similar for bboxes) because:
- `movement` already has I/O parsers/converters for all sorts of formats, so maintaining one more wouldn't be much of a burden for us
- SpikeInterface already has a `to_pynapple_tsgroup()` exporter, and the situation there is analogous. Having the domain-specific packages (SpikeInterface for ephys, `movement` for behaviour) export data towards the package that integrates multiple sources (`pynapple`) makes conceptual sense to me.

  But I'm definitely missing some things here, so let me know if you have arguments for also having a `nap.from_movement` function.

## Thoughts on the routes

I'm not yet sure if we should invest our efforts into the NWB route or the direct route, probably both? The direct route is probably the most flexible/performant, but we would probably also benefit from better integration with the NWB ecosystem. Let me know your thoughts.

Whatever route we take, we should definitely have a docs example on both of our websites, to aid discoverability of the `movement`-to-`pynapple` route.

## Skeletons

I also have some thoughts on skeletons, but will comment on that separately under https://github.com/pynapple-org/pynaviz/pull/118.

lochhh:
Thanks @niksirbi for the write-up. I also want to point out that `movement` has some support for adding Skeletons when saving to NWB (via [`NWBFileSaveConfig`](https://movement.neuroinformatics.dev/latest/api/movement.io.nwb.NWBFileSaveConfig.html#movement.io.nwb.NWBFileSaveConfig.skeleton_kwargs)), so if `pynapple` could read `ndx-pose` objects (or the corresponding `pynwb` parent classes), `pynaviz` could also extract skeleton info from the NWB file created by `movement`.

gviejo:
Dear @niksirbi and @lochhh, thanks for taking the time to look into it. I think both routes make sense to me. So what would need to be done is : 
- the indirect route that requires us to update pynapple to load better ndx-pose
- the direct route could be `movement` exporting a npz file (or directly a pynapple object if you are willing to add `pynapple` as a dependency).  
- an update of documentation on both sides (`pynapple`/`pynaviz` and `movement`).

Does that sound like a good plan?

For us, the cherry on the cake would be an example that mixes tracking data and electrophysiological data.

niksirbi:
Hey @gviejo. 

That plans sounds good!

A clarification on the 2nd point:

> - the direct route could be movement exporting a npz file (or directly a pynapple object if you are willing to add pynapple as a dependency).

I'm currently considering a direct-to-pynapple object export, similar to how our [`to_nwb_file()`](https://movement.neuroinformatics.dev/dev/api/movement.io.save_poses.to_nwb_file.html) function returns an NWB object. We could make `pynapple` an optional dependency (`pip install movement[pynapple]`). If people try to use the direct export function without having `pynapple` installed, we can give them an informative error message.

One caveat for that plan is that we publish `movement` both on PyPI and conda-forge. The latter requires all of our dependencies to also be available on conda-forge, and I think pynapple isn't? Would you in-principle consider [adding `pynapple` to conda-forge](https://conda-forge.org/docs/maintainer/adding_pkgs/)? The process involves several steps, but shouldn't be too hard for a pure Python package like pynapple (I can help if needed). This is more work, but the added benefit is that any package on `conda-forge` could from now on freely depend on `pynapple`. If you don't want to go that way, we will find another way to do the export in `movement` (perhaps directly to an `.npz` file).

I will write up this 'direct route' option as a `movement` issue. 

UPDATE: issue opened here https://github.com/neuroinformatics-unit/movement/issues/1098

gviejo:
Sounds good. We can be on conda-forge. I think as a first pass, `movement` can export to pynapple through npz. That would be the easiest thing to do for now. And when pynapple is on conda-forge, we can revisit the exporter?

niksirbi:
> Sounds good. We can be on conda-forge. I think as a first pass, `movement` can export to pynapple through npz. That would be the easiest thing to do for now. And when pynapple is on conda-forge, we can revisit the exporter?

Agreed!

PR #619, opened 2026-07-21
Speedups of Core Methods

BalzaniEdoardo:
# Speed up time-series reconstruction and `restrict`

## Summary

In this branch we focus on the performance of core `pynapple` methods. One of the main reasons for the slowdowns was the safety checks and data copies performed when creating objects.

These checks are very handy for the user base and are important for pynapple; however, some of them are not always necessary. In particular:

- When internal functions operate on the time axis, we usually know the operation is safe (it preserves temporal ordering, keeps data within the support, etc.), so those checks can be turned off.
- When creating a time series whose support is a single epoch, checking support coverage reduces to testing whether the first and last timestamps fall within the support. If they do, initialization is done; otherwise `restrict` is called.

Additionally, time series creation is now funneled through a single function for consistency.

The mechanism we use to toggle the safety checks is based on a private context manager backed by a `contextvars.ContextVar`. The context is thread- and async-safe and restores its previous state on exit, so the checks are disabled only for the specific internal construction we wrap — never for user-facing construction.

## Changes

**1. Single construction funnel.** Consolidated `_define_instance` so the whole `Tsd`/`TsdFrame`/`TsdTensor`/`Ts` family reconstructs through one chokepoint (`_initialize_tsd_output`) via one bridge (`_ReconstructMixin`). No behavior change; removes a `Ts`-only bypass path.

**2. Trusted-construction fast path.** A private `contextvars` flag (`trusted_construction()`) lets an internal call site assert that the arrays it produced are already float64, sorted, and within-support, so the constructor skips the corresponding revalidation (the O(n) sort-check, the second `_restrict` clip, the unconditional cast). Enabled only where the invariant provably holds: `restrict`, `count`, `value_from`, `in_interval`, `copy`, `time_diff`, `__array_ufunc__`, `__array_function__` (non-time-modifying), `bin_average`, `dropna`, `convolve`, `decimate`, `interpolate`, `derivative`, and order-preserving `__getitem__` (positive-step slices / boolean masks only). **Public construction (`nap.Tsd(...)` from user arrays) is unchanged and still fully validates.**

**3. `searchsorted` `restrict` with dispatch.** For few intervals over numpy data, boundaries are found with `searchsorted` and the selection is copied with plain numpy contiguous slice assignment (one memcpy per interval — this matches a stripped-down slicer on the single-epoch case and beats a fancy-index gather). For many intervals (past ~n/1000) — or non-numpy/lazy data — it falls back to the existing merge scan, which is sequential and cache-friendly. The threshold sits on the measured crossover, so the many-interval case never regresses.

**4. Leaner public construction.** The support-clip in `__init__` now skips entirely when the support already covers all timestamps (the common case) and, when a real clip is needed, rebuilds the index as trusted (the clipped index is a sorted float64 subset). Validation of user input is preserved.

**5. Benchmarks + tests.** Adds `benchmarks/benchmark_core_ops.py` and `tests/test_restrict_routing.py` (spies on the dispatch to assert the intended path is taken per interval count, plus `searchsorted`-vs-scan output equivalence across all types).

## Performance (N = 10M, median/call, JIT warm)

| Operation | before | after | speedup |
|---|---|---|---|
| construct `Tsd` (explicit support) | 37 ms | 7.0 ms | 5.3× |
| `restrict` (1 epoch) | 33 ms | 1.5 ms | 22× |
| `restrict` (100 epochs) | 35 ms | 1.7 ms | 20× |
| `restrict` (1000 epochs) | 34 ms | 4.0 ms | 8× |
| `TsdFrame.restrict` (1 epoch) | 113 ms | 22 ms | 5× |
| `get()` | 15 ms | 12 µs | ~1000× (see note) |
| `count` / `TsGroup.count` | — | unchanged | — |

## ⚠️ Behavior notes for reviewers

Two aliasing/semantics changes fall out of skipping the redundant copy. Both make previously-inconsistent paths consistent with pynapple's **existing** default behavior (`convert_to_numpy_array` already returns ndarrays as-is), but please confirm they're acceptable:

- **Slice / `get()` now return a view**, not a copy — `tsd[a:b]` / `tsd.get(a, b)` share memory with the parent, so in-place writes through them affect the original. Fancy indexing and `restrict` still copy. (Verified: a slice/`get()` result is a genuine view of exactly the requested window.)
- **Construction with an explicit, fully-covering `time_support`** now leaves `values` aliasing the input array (as the default, no-support path already did), instead of the incidental copy the old clip produced.

If copy semantics must be preserved for these, `get()`/slice can force a copy (still ~10× faster than before from skipping the sort-check / re-clip).

## Testing

Full suite green (5215 passed, 274 skipped). New routing tests validate the dispatch and cross-path equivalence.

## Not included (considered, deferred)

- Trusting **public** construction. Should we include a bool flag as a parameter. Something like `skip_checks` or `unsafe` so that people can have a fast init of the time series object if they are confident?

### DIFF
--- a/benchmarks/benchmark_core_ops.py
+++ b/benchmarks/benchmark_core_ops.py
@@ -48,9 +48,12 @@ def main():
     results = {}
     for n in SIZES:
         t, d = make_data(n)
-        support = nap.IntervalSet(float(t[0]), float(t[-1]))
+        # An explicit support covering all timestamps: the common "coverage"
+        # case, and the one that exercises the construction fast path (with
+        # time_support=None the clip is skipped on every version, hiding it).
+        support = nap.IntervalSet(t[0], t[-1])
         tsd = nap.Tsd(t=t, d=d, time_support=support)
-        a, b = float(t[n // 4]), float(t[3 * n // 4])
+        a, b = t[n // 4], t[3 * n // 4]
         one_epoch = nap.IntervalSet(a, b)
         hundred_epochs = many_epochs(t, 100)
         bin_size = (t[-1] - t[0]) / 1000
--- a/benchmarks/benchmark_core_ops.py
+++ b/benchmarks/benchmark_core_ops.py
@@ -1,109 +1,76 @@
-"""Micro-benchmark for pynapple core operations: object construction,
-``restrict`` (single- and many-interval), indexing, and ``count``.
+"""Benchmark core pynapple operations touched by the speedups branch.
 
-Runs on synthetic sorted timestamps of increasing size, and reports the median
-wall time per call (numba JIT warmed up beforehand, so steady-state is measured
-rather than one-off compilation).
-
-Usage::
+Run on the PR branch and on `dev`/`main` to compare:
 
     python benchmarks/benchmark_core_ops.py
-    python benchmarks/benchmark_core_ops.py --sizes 100000 1000000
 
-These numbers back the trusted-construction and searchsorted-``restrict``
-optimizations: reconstruction paths (``restrict``/``get``/``count``/...) skip
-redundant revalidation, and ``restrict`` finds interval boundaries with
-``searchsorted`` + contiguous copies for realistic interval counts, falling back
-to the merge scan when intervals are very numerous.
+Times object construction, `restrict` (one and many epochs), `get`, and
+`count`, at a few data sizes. Numba is warmed up first so steady-state is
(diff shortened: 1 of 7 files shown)


BalzaniEdoardo:
Other thing to notice: I added a benchmarking script but now it is not run anywhere. I think we need to add a docpage dedicated to benchmarking.

gviejo (approved):

PR #634, opened 2026-08-15
Improve tuning curve bin validation

### LINKED ISSUE #629: Better error when passing bins=array for compute_tuning_curves.
The right syntax is bins=[array] but the error is very uninformative.
```
Traceback (most recent call last):
  File "/mnt/home/gviejo/UFOphysio/python/main_pyna_DG_tunings.py", line 192, in <module>
    "ahv" : nap.compute_tuning_curves(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/home/gviejo/pynapple/pynapple/process/tuning_curves.py", line 265, in compute_tuning_curves
    occupancy, bin_edges = np.histogramdd(features, bins=bins, range=range)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/home/gviejo/pynapple/pynapple/core/time_series.py", line 388, in __array_function__
    out = func._implementation(*new_args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/home/gviejo/mambaforge/envs/pynaviz/lib/python3.12/site-packages/numpy/lib/_histograms_impl.py", line 998, in histogramdd
    raise ValueError(
ValueError: The dimension of bins must be equal to the dimension of the sample x.

Process finished with exit code 1
```

AtomicGlance:
## Summary

- validate that `bins` provides one specification per feature before calling NumPy
- explain the required `bins=[bin_edges]` form when explicit one-dimensional edges are passed directly
- cover both the reported single-feature mistake and a general multi-feature count mismatch

Closes #629

## Why

`numpy.histogramdd` currently reports only that the dimensions differ. That does not tell a Pynapple user that explicit edges are expected as one entry per feature. Validating the contract at the API boundary makes the correction immediate while leaving supported scalar, per-feature count, and wrapped edge-array forms unchanged.

## Validation

- focused `test_compute_tuning_curves_type_errors` matrix: 55 passed
- Black: passed
- isort: passed
- Flake8: passed
- Python compilation and `git diff --check`: passed

The full environment could not be installed locally because PyPI was unreachable; the repository CI will run the complete dependency-backed suite.

### DIFF
--- a/pynapple/process/tuning_curves.py
+++ b/pynapple/process/tuning_curves.py
@@ -50,7 +50,8 @@ def compute_tuning_curves(
         The bin specification:
 
         * A sequence of arrays describing the monotonically increasing bin
-          edges along each dimension.
+          edges along each dimension. For a single feature, the edge sequence
+          can be passed directly without wrapping it in another sequence.
         * The number of bins for each dimension (nx, ny, ... =bins)
         * The number of bins for all dimensions (nx=ny=...=bins).
     range : sequence, optional
@@ -222,17 +223,30 @@ def compute_tuning_curves(
             raise ValueError("feature_names should match the number of features.")
 
     # check bins
+    n_features = 1 if features.ndim == 1 else features.shape[1]
+    if (
+        n_features == 1
+        and isinstance(bins, np.ndarray)
+        and bins.ndim == 1
+        and len(bins) > 1
+    ):
+        bins = [bins]
+    elif (
+        n_features == 1
+        and isinstance(bins, (list, tuple))
+        and len(bins) > 1
+        and all(np.isscalar(bin_spec) for bin_spec in bins)
+    ):
+        bins = [bins]
     try:
         n_bin_specs = len(bins)
     except TypeError:
         n_bin_specs = None
-    n_features = 1 if features.ndim == 1 else features.shape[1]
     if n_bin_specs is not None and n_bin_specs != n_features:
         raise ValueError(
             "bins should contain one specification per feature "
             f"(expected {n_features}, got {n_bin_specs}). To use explicit bin "
-            "edges, pass one array per feature (for example, "
-            "bins=[bin_edges] for a single feature)."
+            "edges with multiple features, pass one array per feature."
         )
 
     # check epochs
--- a/pynapple/process/tuning_curves.py
+++ b/pynapple/process/tuning_curves.py
@@ -221,6 +221,20 @@ def compute_tuning_curves(
         ):
             raise ValueError("feature_names should match the number of features.")
 
+    # check bins
+    try:
+        n_bin_specs = len(bins)
+    except TypeError:
+        n_bin_specs = None
+    n_features = 1 if features.ndim == 1 else features.shape[1]
+    if n_bin_specs is not None and n_bin_specs != n_features:
+        raise ValueError(
+            "bins should contain one specification per feature "
+            f"(expected {n_features}, got {n_bin_specs}). To use explicit bin "
+            "edges, pass one array per feature (for example, "
+            "bins=[bin_edges] for a single feature)."
+        )
+
     # check epochs
     if epochs is None:
         epochs = features.time_support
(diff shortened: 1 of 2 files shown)


wulfdewolf:
Hi @AtomicGlance , thanks for that, looking good!

It would actually be ideal if we added the option to pass bins without wrapping the list around it, if and only if there is just one feature.
Would you be willing to add that?
It might require some extra checks and tests.

AtomicGlance:
> Hi @AtomicGlance , thanks for that, looking good!
> 
> It would actually be ideal if we added the option to pass bins without wrapping the list around it, if and only if there is just one feature. Would you be willing to add that? It might require some extra checks and tests.

Hi, just implemented those check them out and tell me if it's alright.

pynapple/process/tuning_curves.py
```diff
@@ -221,6 +222,33 @@ def compute_tuning_curves(
...
+        n_bin_specs = len(bins)
+    except TypeError:
+        n_bin_specs = None
+    if n_bin_specs is not None and n_bin_specs != n_features:
+        raise ValueError(
+            "bins should contain one specification per feature "
+            f"(expected {n_features}, got {n_bin_specs}). To use explicit bin "
+            "edges with multiple features, pass one array per feature."
+        )
+
```
wulfdewolf, line 239:
```suggestion
# check bins
n_features = 1 if features.ndim == 1 else features.shape[1]

# For a single feature, a 1D sequence is interpreted as explicit
# bin edges and does not need to be wrapped by the caller.
if n_features == 1 and np.ndim(bins) == 1:
    bins = [bins]

try:
    n_bin_specs = len(bins)
except TypeError:
    n_bin_specs = None

if n_bin_specs is not None and n_bin_specs != n_features:
    raise ValueError(
        "bins should contain one specification per feature "
        f"(expected {n_features}, got {n_bin_specs}). To use explicit bin "
        "edges with multiple features, pass one array per feature."
    )
```

wulfdewolf: I think using `np.ndim` here can make this a lot shorter

tests/test_tuning_curves.py
```diff
@@ -277,6 +277,31 @@ def get_features_n(n, fs=10.0):
...
+        pytest.param(
+            get_group_n(1),
+            get_features_n(2),
+            {"bins": [5]},
+            pytest.raises(
+                ValueError,
+                match=r"bins should contain one specification per feature \(expected 2, got 1\)",
+            ),
+            id="bins-wrong-feature-count",
+        ),
```
wulfdewolf, line 290: Since we are now checking `bins` ourselves (as opposed to offloading to numpy before), we should check all possible input formats.

wulfdewolf (commented):
Looking good, I suggested the use of `np.ndim` to make the check shorter.
I think we should now also test all possible input formats, so a couple more tests are needed.

AtomicGlance:
Thanks — I pushed 2677418 with the requested cleanup. Single-feature bin normalization now uses np.ndim, and the direct-bin test covers NumPy arrays, lists, and tuples. The focused tuning-curve suite passes locally (237 passed). Could you take another look when you have a chance?

wulfdewolf:
Hi @AtomicGlance , thanks for that!
I'm going to be nitpicky and ask for more, apologies.
Since we are now not offloading the checks for `bins` to `np.histogramdd` anymore, I really want all possible input types tested, for example:
- a single integer as number of bins for all features, or for just one
- a list of integers as numbers of bins for all features
- a list of edges
- ...

The check itself looks fine now!

AtomicGlance:
Added the full `bins` input matrix in c367a0c.

The tests now cover scalar counts for one or every feature, per-feature count sequences as lists, tuples, and NumPy arrays, mixed count/edge specifications, direct and wrapped single-feature edges, and multi-feature edge sequences as lists, tuples, and arrays. The existing wrong-feature-count case remains covered.

Validation: 247 tuning-curve tests pass; Ruff lint and formatting checks pass. This should exercise each branch introduced by the local bin normalization.

GitHub has gated the new CI and Codespell runs (`32158210593` and `32158210566`) before starting any jobs. @wulfdewolf, could you approve them and take another look when they pass?

gviejo:
Thanks @AtomicGlance merging it.

PR #640, opened 2026-08-28
Add lagged cross-correlation for TsdFrame

### LINKED ISSUE #613: Implement time lagged correlation of TsdFrame
Something like `nap.compute_lagged_crosscorrelation(tsdframe, window)`

viranovskaya:
Closes #613.

This adds `compute_lagged_crosscorrelation` for regularly sampled `TsdFrame` data. With one frame, it computes all unique signal pairs; with two frames, it computes all cross-frame pairs.

Observations are pooled across epochs without crossing epoch boundaries. Positive lags indicate that the second signal follows the first.

The calculation uses a JIT-compiled online Pearson update, with the sampling interval taken from the timestamps. Tests cover NumPy equivalence, shifted signals, multiple epochs, numerical stability, invalid inputs and edge cases. The user guide is updated as well.

### DIFF
--- a/doc/user_guide/05_correlograms_isi.md
+++ b/doc/user_guide/05_correlograms_isi.md
@@ -67,6 +67,42 @@ print(crosscorrs)
 
 Column name `(0, 1)` is read as cross-correlogram of neuron 0 and 1 with neuron 0 being the reference time.
 
+## Lagged cross-correlations
+
+Lagged cross-correlation measures how continuous signals vary together at different time offsets. The input is a regularly sampled `TsdFrame`; no binning is needed because its columns are already sampled on the same time axis.
+
+```{code-cell} ipython3
+rng = np.random.default_rng(1)
+timestamps = np.arange(1000) / 100
+reference = rng.normal(size=len(timestamps))
+delayed = np.r_[np.zeros(10), reference[:-10]]
+signals = nap.TsdFrame(
+    t=timestamps,
+    d=np.column_stack((reference, delayed)),
+    columns=["reference", "delayed"],
+    time_support=nap.IntervalSet(0, 10),
+)
+
+lagged_corr = nap.compute_lagged_crosscorrelation(
+    signals, windowsize=0.5, time_units="s"
+)
+print(lagged_corr)
+```
+
+The output is a pandas DataFrame with lags in seconds as its index and signal pairs as its columns. Here the correlation peaks at `+0.1` seconds: positive lags mean that the second signal follows the first.
+
+Passing two `TsdFrame` objects computes all cross-frame column pairs. Their timestamps must match.
+
+```{code-cell} ipython3
+reference_frame = signals[:, [0]]
+delayed_frame = signals[:, [1]]
+cross_frame_corr = nap.compute_lagged_crosscorrelation(
+    (reference_frame, delayed_frame), windowsize=0.5
+)
+```
+
+The optional `epochs` argument restricts the calculation. Valid observations are pooled across epochs for each lag, but pairs never cross an epoch boundary.
+
 ## Event-correlograms
 
 Event-correlograms count the number of event in the `TsGroup` based on an `event` timestamps object. 
--- a/pynapple/process/correlograms.py
+++ b/pynapple/process/correlograms.py
@@ -161,6 +161,224 @@ def _cross_correlogram(
     return C, B
 
 
+@jit(nopython=True, cache=True)
+def _lagged_crosscorrelation(
+    data1: npt.NDArray[np.float64],
+    data2: npt.NDArray[np.float64],
+    counts: npt.NDArray[np.int64],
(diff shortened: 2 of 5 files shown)


doc/user_guide/05_correlograms_isi.md
```diff
@@ -67,6 +67,42 @@ print(crosscorrs)
 
 Column name `(0, 1)` is read as cross-correlogram of neuron 0 and 1 with neuron 0 being the reference time.
 
+## Lagged cross-correlations
+
+Lagged cross-correlation measures how continuous signals vary together at different time offsets. The input is a regularly sampled `TsdFrame`; no binning is needed because its columns are already sampled on the same time axis.
```
wulfdewolf: 'no binning is needed because its columns are already sampled on the same time axis.` can be dropped

doc/user_guide/05_correlograms_isi.md
```diff
@@ -67,6 +67,42 @@ print(crosscorrs)
...
+    time_support=nap.IntervalSet(0, 10),
+)
+
+lagged_corr = nap.compute_lagged_crosscorrelation(
+    signals, windowsize=0.5, time_units="s"
+)
+print(lagged_corr)
+```
+
+The output is a pandas DataFrame with lags in seconds as its index and signal pairs as its columns. Here the correlation peaks at `+0.1` seconds: positive lags mean that the second signal follows the first.
```
wulfdewolf, line 92: We definitely want a TsdFrame as output!

pynapple/process/correlograms.py
```diff
@@ -161,6 +161,224 @@ def _cross_correlogram(
...
+        raise TypeError("windowsize must be a number.")
+    if not np.isfinite(windowsize) or windowsize < 0:
+        raise ValueError("windowsize must be finite and non-negative.")
+    if epochs is not None and not isinstance(epochs, nap.IntervalSet):
+        raise TypeError("epochs must be an IntervalSet or None.")
+    if not isinstance(time_units, str):
+        raise TypeError("time_units must be a string.")
+
+    if np.iscomplexobj(data1.values) or np.iscomplexobj(data2.values):
+        raise TypeError("Lagged cross-correlation requires real-valued data.")
```
wulfdewolf, line 354: this can be dropped, should be handled by `TsdFrame` already

doc/user_guide/05_correlograms_isi.md
wulfdewolf: Hmm scratch that for now, we can't yet handle a multiindex in a `TsdFrame`, we're discussing whether we should.
For now the `DataFrame` output is fine.

tests/test_correlograms.py
```diff
@@ -193,6 +193,341 @@ def test_correlograms_type_errors_event(func, args, msg):
...
+
+
+def test_lagged_crosscorrelation_rejects_complex_data():
+    frame = nap.TsdFrame(
+        t=np.arange(5),
+        d=np.column_stack((np.arange(5), np.arange(5))) * (1 + 1j),
+    )
+
+    with pytest.raises(TypeError, match="real-valued data"):
+        nap.compute_lagged_crosscorrelation(frame, 1)
```
wulfdewolf: can also be dropped

wulfdewolf (commented):
Hi @viranovskaya ,

Thanks for this! It looks good already.
I added a couple small comments, mostly about unneeded text or checks.

We are discussing whether we can allow a multiindex in `TsdFrame`, which could allow this function to return a `TsdFrame` instead of a `pd.DataFrame`.
For now the `pd.DataFrame` is fine though.

One thing I am uncertain about is what happens when there are nans in the continuous traces.
Can we document that in the docstring and state explicitly what happens?

viranovskaya:
Thanks for the review! I’ve shortened the user-guide sentence and added a Notes section to the docstring explaining the NaN behavior explicitly.

NaNs are propagated, not omitted or interpolated. For each column pair and lag, the correlation is NaN if either signal contains a NaN among the overlapping observations pooled across the selected epochs. Pairs not involving the NaN-containing columns are unaffected, as are observations outside the selected epochs or lag overlap. I’ve added regression tests covering these cases for both single- and two-frame inputs.

The DataFrame output is unchanged.

I checked the complex-input validation before removing it. `TsdFrame` supports complex data, including the documented output of `apply_hilbert_transform`. This function computes real-valued Pearson correlations; without the guard, the float64 conversion drops the imaginary component with a `ComplexWarning`. I’ve therefore kept the explicit rejection and extended the tests to cover complex data in either input frame. Does keeping that function-specific check make sense here?

Locally, the full test suite passes: 5,330 passed and 274 skipped. The correlogram doctests and lint checks also pass. The full doctest run has one float-representation failure in `apply_hilbert_transform`, which I reproduced on the unmodified PR head in the same environment.

PR #641, opened 2026-08-31
DOC: add instruction for uv to CONTRIBUTING.md

just-meng:
that's what i do, feel free to say no if not relevant :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,13 @@ conda activate pynapple
 pip install -e ".[dev,docs]"
 ```
 
+Alternatively, [uv](https://docs.astral.sh/uv/) creates the virtual environment and installs pynapple in editable mode in a single step:
+
+```bash
+uv sync --extra dev --extra doc --python=3.8
+source .venv/bin/activate
+```
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -19,6 +19,13 @@ conda activate pynapple
 pip install -e ".[dev,docs]"
 ```
 
+Alternatively, [uv](https://docs.astral.sh/uv/) creates the virtual environment and installs pynapple in editable mode in a single step:
```
gviejo, line 22: I am not familiar with uv. Do you have to be in the pynapple folder for this to work?

just-meng: Yes, `uv sync` looks for `pyproject.toml` in the current directory. Alternatively, `uv sync --project /path/to/pynapple` works from anywhere.

just-meng: I think uv is gaining much popularity since it is much faster and smarter than conda and offers a range of provenance-tracking options for computational environments. Might be worth adding it to this document.

gviejo: Thanks Meng merging it.

gviejo (approved):

Issue #644, opened 2026-08-31
TsdFrame accepts tuple/MultiIndex columns but raises on any column access


just-meng:
`TsdFrame` accepts a pandas `MultiIndex` (or any tuple labels) as `columns` at construction, silently flattens it to a plain `Index` of tuples, and then raises on every column access.

## Reproducing

```python
import numpy as np
import pandas as pd

import pynapple as nap

df = pd.DataFrame(
    {
        "time": np.tile(np.arange(5), 4),
        "region": np.repeat(["ca1", "ca1", "ca3", "ca3"], 5),
        "layer": np.repeat([1, 2, 1, 2], 5),
        "value": np.arange(20, dtype=float),
    }
)
wide = df.pivot_table(index="time", columns=["region", "layer"], values="value")
print(type(wide.columns))
print(wide[("ca1", 1)].tolist())

tsdframe = nap.TsdFrame(t=wide.index.values, d=wide.values, columns=wide.columns)
print(type(tsdframe.columns))
print(list(tsdframe.columns))

tsdframe[:, 0] 

# or this variant:
tsdframe.loc[("ca1", 1)]    # IndexError: 1
```

Output:

```
<class 'pandas.MultiIndex'>
[0.0, 1.0, 2.0, 3.0, 4.0]
<class 'pandas.Index'>
[('ca1', 1), ('ca1', 2), ('ca3', 1), ('ca3', 2)]
Traceback (most recent call last):
  File "repro.py", line 26, in <module>
    tsdframe[:, 0]
    ~~~~~~~~^^^^^^
  File "pynapple/core/metadata_class.py", line 56, in _decorator
    return func(self, *args, **kwargs)
  File "pynapple/core/time_series.py", line 1968, in __getitem__
    kwargs["metadata"] = self._metadata.loc[columns]
  File "pynapple/core/metadata_class.py", line 842, in __getitem__
    idx = self.index_map[index]  # indexer from index map
KeyError: 'ca1'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/Data/repos/pynapple/pynapple/core/utils.py", line 339, in __getitem__
    raise IndexError(str(k))
IndexError: ca1
```

## What is happening

Two separate things:

1. The `MultiIndex` is not preserved. `type(tsdframe.columns)` is `Index`, not `MultiIndex`, so the column labels degrade into opaque tuples.
[...]

PR #654, opened 2026-09-11
Fix positional TsdFrame selection with tuple column labels

GN-Yu:
`frame[:, 0]` raises `KeyError` or `IndexError` when a `TsdFrame` column has a
tuple label, including labels supplied through a pandas `MultiIndex`. The data
selection succeeds, but the internal metadata lookup interprets the selected
tuple as separate row and column indexers.

Wrap a selected tuple label in a one-element list for that metadata lookup. This
allows scalar positional selection to return the expected `Tsd` and preserves
existing list/slice selection behavior.

Addresses the scalar positional-selection behavior discussed in #644. Full
MultiIndex preservation and public tuple-based `.loc` semantics remain separate.

Regression tests cover two- and three-part tuples, MultiIndex input, metadata
present/absent, Python/NumPy integers, negative indices, and list/slice controls.
They check values, timestamps, nontrivial time support, return types, and column
and metadata alignment for frame outputs.

Validation on Linux/WSL, Python 3.11, NumPy/Numba backend:

- New tests before the fix: 24 failed, 24 passed.
- New tests after the fix: 48 passed.
- Full test suite under coverage: 5,398 passed, 274 skipped.
- Core/process doctests: 91 passed, 1 skipped.
- Black, isort, Flake8, and `git diff --check` passed.

Implementation and regression tests were prepared with Codex assistance and
validated locally.


gviejo:
Looks good. Thanks.

PR #656, opened 2026-09-13
Fix ISI histogram bins for constant intervals

viranovskaya:
I traced the intermittent ISI test failure in #640 to the existing bin-edge calculation. When all intervals are equal, `linspace(min_isi, max_isi, bins + 1)` produces repeated edges. This puts every observation in the last bin and gives duplicate output indices.

This uses `np.histogram_bin_edges` for integer bins, matching NumPy's handling of a constant range while retaining shared edges for a TsGroup. Explicit bin edges are unchanged.

The regression tests cover a single interval and repeated equal intervals across Ts, Tsd, TsdFrame, TsdTensor and TsGroup, with linear and log scales. The grouped test reference now uses NumPy's histogram edges too.

Checked locally on current dev: 5,390 unit tests passed, 274 skipped; Black, isort and flake8 passed. The failing implementation was identical in the base and head of #640, so I have kept this fix separate from that feature.


pynapple/process/correlograms.py
BalzaniEdoardo: if the bins are always equispaced, we could use arithmetic for index assignment and then use `np.bincount` to get the histogram, and it should be faster than repeated histogram calls:

```python

bin_min = min_isi
bin_max = max_isi
bin_width = bin_edges[1] - bin_edges[0]
num_bins = len(bin_edges) - 1


hist_dict = {}
for i, series in time_diffs.items():
    # Convert to a raw numpy array of values
    vals = series.values
    
    # Map values to integer bin indices: floor((val - min) / width)
    bin_indices = ((vals - bin_min) / bin_width).astype(np.intp)
    
    # Handle values outside your fixed bin edges (clip them to an extra boundary or discard)
    # This ensures everything stays within [0, num_bins - 1]
    valid_mask = (vals >= bin_min) & (vals < bin_max)
    valid_indices = bin_indices[valid_mask]
    
    # Count occurrences instantly using bincount
    hist_dict[i] = np.bincount(valid_indices, minlength=num_bins)

```

this should be tested and benchmarked on large time series to check if the approach is actually better

gviejo (approved):

gviejo:
Merging it. Thanks @viranovskaya

PR #653, opened 2026-09-10
BF: Fix TsdFrame groups collapsing to Tsd for single-column groups

just-meng:
Closes #643.

Rewritten after @gviejo's review: instead of routing `groupby` around `.loc`, this fixes `.loc` itself, at the line pointed out in the review.

Changes to the implementation:
`_TsdFrameSliceHelper.__getitem__` (utils.py) resolved the key to column positions and then chose the return shape with `if len(index) == 1`. A scalar key always resolves to one position, but so does a list of one label, so `tsdframe.loc[["a"]]` and `tsdframe[["a"]]` returned a Tsd while `tsdframe[:, [0]]` returned a TsdFrame. It now decides from the key, as pandas does: a scalar key drops the column dimension, a list-like key keeps it whatever it selects. `groupby` indexes its groups with a list of labels, so single-column groups stop collapsing as a side effect.

Changes to the tests include:
- a contract test `test_tsdframe_column_selection`, over `loc`, plain indexing and positional selection, for scalar and list keys
- a regression test `test_groupby_apply_column_reduction_on_single_column_group` for the reported symptom
- modification of the existing fixture `tsdframe_gba` to include single-column groups
- modification of `test_metadata_groupby` and `test_metadata_groupby_apply_func_kwargs`, which skipped the collapsing case and compared against `.loc` itself

The `TsdFrame.groupby_apply` doctest had pasted the collapsed output as its expected result and is corrected.

Drafted by Claude Code, reviewed & finalized by me.

PR #660, opened 2026-09-22
Fix TsdFrame.loc reading a composite column label as a list of labels

just-meng:
Follow-up to #653, as discussed there. Partially addresses #644.

`_TsdFrameSliceHelper.__getitem__` treats any iterable key as a collection of column labels, so a tuple label such as `("ca1", 1)` was iterated into `"ca1"` and `1` and raised `IndexError`, although it is a single column of a `TsdFrame` built from tuples or from a pandas `MultiIndex`.

The behaviour you asked for, all four now:

```python
tsdframe.loc[("ca1", 1)]              # -> Tsd (5,)        was IndexError
tsdframe.loc["ca1", 1]                # -> Tsd (5,)        same call, Python builds the tuple
tsdframe.loc[[("ca1", 1)]]            # -> TsdFrame (5, 1) already worked after #653
tsdframe.loc[[("ca1", 1), ("ca3", 2)]] # -> TsdFrame (5, 2) already worked after #653
```

The fix takes the single-label path when the key is itself one of the columns. A tuple that is not a column keeps its previous meaning, a collection of labels, which is what makes `tsdframe.loc[("ca1", 1), ("ca3", 1)]` select two columns.

Tests: `test_tsdframe_column_selection` gains a row for tuple labels rather than a test of its own, since the rule is the same one it already states, that a scalar key drops the column dimension and a list key keeps it. One of its eight cases fails without this change.

Still open in #644:

- `tsdframe[("ca1", 1)]` cannot work: for `__getitem__` a tuple key is the two-axis form, `tsdframe[:, 0]` being `tsdframe[(slice(None), 0)]`, so a tuple there means `(rows, columns)`. `loc` has no such conflict, which is why all four cases above are `loc`.
- `tsdframe[[("ca1", 1)]]` could work but does not. The convenience layer in `TsdFrame.__getitem__` routes a list to `loc` only when every element is a string.
- A `pandas.MultiIndex` passed as `columns` still degrades to an `Index` of tuples at construction.
- A mistyped composite label reports the offending element rather than the label, `tsdframe.loc[("ca1", 99)]` raising `IndexError: 99`, because by then the key is being read as a collection.

Drafted by Claude Code, reviewed & finalized by me.


gviejo:
Good for me. I don't think this behavior `tsdframe[("ca1", 1)]` should work. The behavior of pynapple should be closer to numpy than to pandas. Merging it.

